Issue #2, opened 2024-09-12
Define basic repo workflows


phbnf:
- Submit configs for
  - [x] dependabot
  - [x] codeql: it's not in the trillian-tessera repo, but it's in the c-t-go repo, let's add it here as well and see how it goes
  - [x] go_test
  - [x] golanci-lint
  - [x] govulncheck
 
 - Edit the repo config such that it requires these workflows to pass:
   - [x] go_test
   - [x] golanci-lint: we don't have this requirement in trillian-tessera, but I think it should be fine to have it? I'll try!
   - [x] govulncheck

Scorecard is enabled in c-t-go, not in trillian-tessera. I'm not enabling it here either since this is still under heavy development. We'll add this later.

roger2hk:
I don't have access to add golangci-lint into the required tests. It should be required.

PR #3, opened 2024-09-12
Define some workflows.

phbnf:
Towards #2 

This PR sets up:
 - dependabot
 - codeql: it's not in the trillian-tessera repo, but it's in the c-t-go repo, let's see how it goes
 - go_test: I'm leaving the `-race` option, it can't hurt.
 - golanci-lint: it's setting the version to v1.61.0 (vs 1.55.X in c-t-go, 1.59.X in trillian-tessera). It passes tests with this last version, so we might as well use it directly. Let's talk more on how to manage version update of this on our repo in our next team meeting.
 - govulncheck

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,21 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: weekly
+  groups:
+    all-deps:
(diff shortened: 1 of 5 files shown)


phbnf:
(Tests fail because I have not submitted code yet, that will come after, and should just work out of the box, 🤞 )

AlCutter (approved):

PR #4, opened 2024-09-12
Import code from transparency-dev/trillian-tessera

phbnf:
I'm moving the modules as well since they are only used by this personality for now and they are under the sctfe directory. We could/should make them available in the trillian-tessera repo in a better directory, but let's handle this after this PR.


AlCutter (approved):
Going to mostly just trust you that this is the same code :)

Definitely in follow-ups rather than here, but can we add tasks in transparency-dev/trillian-tessera#88 to:
- apply standard Go repo structure here (e.g. `cmd/...` for the binaries)
- move ~as much as possible of everything else under `/internal` to reduce public API surface (and we'll consciously expose bits as necessary later on)

phbnf:
Sure thing for the followup todos.

> Going to mostly just trust you that this is the same code :)
Some of the cp commit had been mixed with the sed commit. I've re-arranged everything properly to have a cp commit, and a sed commit. I'll merge them *without* squashing them. If anybody wants to check that it's the same code, run the following in the top level directory of this repo, and in transparency-dev/trillian-tessera/personalities/sctfe: `find ./ -type f -name "*.go" -print0 | sort -z | xargs -0 -I % sha256sum % | sha256sum`

For the record:

```
 [11:24][~/git/phbnf/static-ct] - (HEAD detached at 177d8c4)
└─> find ./ -type f -name "*.go" -print0 | sort -z | xargs -0 -I % sha256sum % | sha256sum
2308a3b98f59c47350f3973a03af2e8035f77518c2d8b65f98b8b81ae807f221  -

┌[11:23][~/git/phbnf/trillian-tessera/personalities/sctfe] - main
└─> find ./ -type f -name "*.go" -print0 | sort -z | xargs -0 -I % sha256sum % | sha256sum
2308a3b98f59c47350f3973a03af2e8035f77518c2d8b65f98b8b81ae807f221  -
```

Issue #5, opened 2024-09-13
Simplify internals of the SCTFE config


phbnf:
There's a bunch of struct layers inside the SCTFE at the moment, let's simplify this. The first layer of complexity, the protobuf config was removed in [trillian-tessera#200](https://github.com/transparency-dev/trillian-tessera/issues/200). This Issue is a continuation of this one.

`ValidateConfig` takes a lot of flags as arguments, and validated them. This should probably be move to main.
Right now, it also parses the ExtendedKeyUsage, we might want to move this around.

Then there's `SetUpInstance` which actually applies some logic on stuff and open the root_pem_file, or parses the oids. Logically, I think that parsing the oids is similar to parsing the ExtendedKeyUsage, so I might group them together.

Also, right now, the Tessera storage cannot be created anywhere else than in `SetupInstance` because, it needs a signer, and the signer is created in there... because we need a note.Signer for the checkpoint, but not for the SCTs.

Then theres's `newLogInfo`, and this one is a bit annoying as well because it's the one that as the TimeSource object, which is also used in tests.

PR #33, opened 2024-10-30
Remove TLS support

phbnf:
Towards #10 #5 

We don't need this for now. It was only added recently to the CTFE and got ported over here. In an effort to splify this binary to its bare minimum, let's remove this for now. We can always add it back later if required.

### DIFF
--- a/cmd/gcp/main.go
+++ b/cmd/gcp/main.go
@@ -17,7 +17,6 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


cmd/gcp/main.go
```diff
@@ -194,11 +178,7 @@ func main() {
 		klog.Info("HTTP server shutdown")
 	})
 
-	if *tlsCert != "" && *tlsKey != "" {
-		err = srv.ListenAndServeTLS("", "")
-	} else {
-		err = srv.ListenAndServe()
-	}
+	err = srv.ListenAndServe()
 	if err != http.ErrServerClosed {
```
roger2hk, line 182: nit: Merge these two lines into one.

roger2hk (approved):

PR #40, opened 2024-11-07
Support GCP Secret Manager for signer key pair

roger2hk:
Towards https://github.com/transparency-dev/static-ct/issues/6.

This pull request adds the GCP Secret Manager for storing the signer key pair. The terraform configs are included for the GCP testing environment.

Note that only ECDSA key is supported. RSA key is NOT supported.

PR #51, opened 2024-11-12
Refactor terraform modules

roger2hk:
This is a follow-up pull request to address https://github.com/transparency-dev/static-ct/pull/40#discussion_r1834174366.


deployment/modules/gcp/main.tf
mhutchinson, line 1: I don't think putting this module at the root is the cleanest encapsulation. What happens when we define a kms module and want a version that is storage+kms ?

Better options:
  1) make the key module the root, and just include storage from there. It seems unlikely you'd want to deploy just the key part on its own
  2) move this module to modules/gcp/full_stack_secretmanager

deployment/modules/gcp/key/main.tf
mhutchinson, line 1: secretmanager is probably a better name than key? If we define another key storage mechanism, the naming falls apart.

deployment/modules/gcp/key/main.tf
```diff
@@ -0,0 +1,60 @@
...
+}
+
+# Secret Manager
+
+# ECDSA key with P256 elliptic curve. Do NOT use this in production environment.
+#
+# Security Notice
+# The private key generated by this resource will be stored unencrypted in your 
+# Terraform state file. Use of this resource for production deployments is not 
+# recommended. Instead, generate a private key file outside of Terraform and
```
phbnf, line 17: Nit - I'd remove the "instead...", let's not make security recommendations until we've figured out what we want to do :)

+1 for the Security notice, thanks!

deployment/modules/gcp/main.tf
phbnf: +1, let's not put this at root, especially since this storage+secretmanager composition is not meant for production. If you want to leave both module separate like this, that's fine by me, but I'd put the composition under /test to make it explicit.

deployment/modules/gcp/key/main.tf
roger2hk: Updated to `secretmanager`.

deployment/modules/gcp/main.tf
roger2hk: I've moved the root module to `conformance` module which will use `storage` and `secretmanager` modules. This `conformance` module will be used in the upcoming CI env.

phbnf (approved):

mhutchinson (approved):

PR #52, opened 2024-11-13
Add missing secretmanager project service and dependencies

roger2hk:
#6


AlCutter (approved):
Feels like this `storage` module name is becoming increasingly inaccurate though, might be worth a TODO to have a ponder on renaming it at some point in the future?

AlCutter:
> Feels like this `storage` module name is becoming increasingly inaccurate though, might be worth a TODO to have a ponder on renaming it at some point in the future?

Ah, that looks like what #51 is about - great minds!
Carry on :)

phbnf (approved):

Issue #53, opened 2024-11-17
[GCP] Setup CI Environment


roger2hk:
When there is a commit merged into the main branch, it will trigger a build and deploy to the CI environment.

PR #63, opened 2024-11-25
Enable Artifact Registry in GCP CI Conformance Terraform

roger2hk:
Towards #53.

This PR implements the TODO for "deployment/live/gcp/ci/", so the Artifact Registry does not need to be created externally.

### DIFF
--- a/deployment/live/gcp/ci/terragrunt.hcl
+++ b/deployment/live/gcp/ci/terragrunt.hcl
@@ -4,6 +4,7 @@ terraform {
 
 locals {
   env                 = "ci"
+  docker_env          = local.env
   base_name           = "${local.env}-conformance"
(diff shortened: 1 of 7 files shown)


deployment/live/gcp/ci/terragrunt.hcl
```diff
@@ -4,6 +4,7 @@ terraform {
 
 locals {
   env                 = "ci"
+  docker_env          = local.env
```
phbnf, line 7: Could you use env directly?

roger2hk: `docker_env` is required as the module is shared in both conformance CI (`env = "ci"`) and cloudbuild (`env = "prod"`). Keeping the `docker_env` is easier to understand at the moment.

PR #64, opened 2024-11-25
Support conformance env lifecycle in GCP CI Cloud Build

roger2hk:
Towards #53.

### DIFF
--- a/deployment/modules/gcp/cloudbuild/main.tf
+++ b/deployment/modules/gcp/cloudbuild/main.tf
@@ -161,7 +161,7 @@ resource "google_cloudbuild_trigger" "build_trigger" {
         openssl x509 -req -days 3650 -in /tmp/httpschain/cert.csr -CAkey testdata/fake-ca.privkey.pem -CA testdata/fake-ca.cert -passin pass:"gently" -outform pem -out /tmp/httpschain/chain.pem -provider legacy -provider default
         cat testdata/fake-ca.cert >> /tmp/httpschain/chain.pem
         cat /tmp/httpschain/chain.pem | jq --raw-input --slurp --compact-output 'split("\n-----END CERTIFICATE-----\n") | map(select(length > 0) | sub("^-----BEGIN CERTIFICATE-----\n"; "") | sub("\n-----END CERTIFICATE-----$"; "")) | { "chain": . }' > /tmp/httpschain/chain.json
-        curl -s -o >(cat > /tmp/add_chain_response_body) -w "%{http_code}" -X POST --data @/tmp/httpschain/chain.json -H "Content-Type: application/json" -H "Authorization: Bearer $(cat /workspace/cb_identity)" $(cat /workspace/conformance_url)/ci-${var.project_id}/ct/v1/add-chain > /tmp/add_chain_response_code
+        curl -s -o >(cat > /tmp/add_chain_response_body) -w "%%{http_code}" -X POST --data @/tmp/httpschain/chain.json -H "Content-Type: application/json" -H "Authorization: Bearer $(cat /workspace/cb_identity)" $(cat /workspace/conformance_url)/ci-${var.project_id}/ct/v1/add-chain > /tmp/add_chain_response_code
 
         cat /tmp/add_chain_response_code
         if ! grep -q 200 /tmp/add_chain_response_code; then
--- a/deployment/modules/gcp/cloudbuild/main.tf
+++ b/deployment/modules/gcp/cloudbuild/main.tf
@@ -31,7 +31,14 @@ resource "google_project_service" "cloudbuild_api" {
   disable_on_destroy = false
 }
 
-## Service usage API is required for roles/serviceusage.serviceUsageViewer.
+## Service usage API is required on the project to enable APIs.
+## https://cloud.google.com/apis/docs/getting-started#enabling_apis
+## serviceusage.googleapis.com acts as a central point for managing the API 
+## lifecycle within your project. By ensuring the required APIs are enabled 
+## and accessible, it allows Cloud Build to function seamlessly and interact 
+## with other Google Cloud services as needed.
+## 
+## The Cloud Build service account also needs roles/serviceusage.serviceUsageViewer.
 resource "google_project_service" "serviceusage_api" {
   service            = "serviceusage.googleapis.com"
   disable_on_destroy = false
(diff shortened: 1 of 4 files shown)


deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -30,41 +31,14 @@ resource "google_project_service" "cloudbuild_api" {
...
-  role    = "roles/run.developer"
-  member  = "serviceAccount:${google_service_account.cloudbuild_service_account.email}"
-}
-
-resource "google_project_iam_member" "iam_service_account_user" {
-  project = var.project_id
-  role    = "roles/iam.serviceAccountUser"
-  member  = "serviceAccount:${google_service_account.cloudbuild_service_account.email}"
+resource "google_project_service" "serviceusage_api" {
+  service            = "serviceusage.googleapis.com"
```
phbnf, line 35: What's that for?

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -119,25 +107,67 @@ resource "google_cloudbuild_trigger" "build_trigger" {
       wait_for = ["docker_build_conformance_gcp"]
```
phbnf, line 107: Should it also wait for preclean_env?

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -119,25 +107,67 @@ resource "google_cloudbuild_trigger" "build_trigger" {
...
+        terragrunt --terragrunt-no-color output --raw conformance_url -no-color > /workspace/conformance_url
+      EOT
+      dir    = "deployment/live/gcp/ci"
+      env = [
+        "GOOGLE_PROJECT=${var.project_id}",
+        "TF_IN_AUTOMATION=1",
+        "TF_INPUT=false",
+        "TF_VAR_project_id=${var.project_id}"
       ]
       wait_for = ["docker_push_conformance_gcp"]
```
phbnf, line 127: Should it also wait for preclean_env?

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -119,25 +107,67 @@ resource "google_cloudbuild_trigger" "build_trigger" {
...
+      wait_for = ["terraform_apply_conformance_ci"]
+    }
+
+    ## Test against the conformance server.
+    ## TODO: Replace this with CT Hammer when it is ready.
+    step {
+      id       = "curl_test"
+      name     = "curlimages/curl"
+      script   = <<EOT
+        curl -iX POST $(cat /workspace/conformance_url)/ci-${var.project_id}/ct/v1/add-pre-chain -H "Authorization: Bearer $(cat /workspace/cb_identity)"
```
phbnf, line 148: Let's generate a chain certificate with openssl to start with and/or store a chain in a test directory? there might already be one over there, we just need to be mindful of notBefore/notAfter.

roger2hk: Done but I think it's not worth to invest time on this as we're going to replace this with CT Hammer soon.

deployment/modules/gcp/cloudbuild/main.tf
roger2hk: Good catch. Added.

deployment/modules/gcp/cloudbuild/main.tf
roger2hk: Added the comment.

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -21,7 +21,8 @@ module "artifactregistry" {
 # Cloud Build
 
 locals {
-  artifact_repo                = "${var.location}-docker.pkg.dev/${var.project_id}/${google_artifact_registry_repository.docker.name}"
+  cloudbuild_service_account   = "cloudbuild-${var.env}-sa@${var.project_id}.iam.gserviceaccount.com"
```
phbnf, line 24: Nit, maybe after this PR is merged - I would recommend passing value of entities created outside of this terraform via variables from Terragrunt directly. That way, it's easier for folks to re-use what we have with their own users. 

Even in Terragrunt, if you use some logic to build this name, people we'd need to use the same conventions for it to work. Hence, we need to explain what the naming convention is. Alternatively, you could just hardcode the full users, rather than building them programatically. We only have one environment for now, so that's not an issue.

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -116,28 +106,77 @@ resource "google_cloudbuild_trigger" "build_trigger" {
...
+      id       = "curl_test"
+      name     = "gcr.io/cloud-builders/gcloud"
+      script   = <<EOT
+        apt update && apt install jq -y
+        mkdir -p /tmp/httpschain
+        openssl genrsa -out /tmp/httpschain/cert.key 2048
+        openssl req -new -key /tmp/httpschain/cert.key -out /tmp/httpschain/cert.csr -config=testdata/fake-ca.cfg
+        openssl x509 -req -days 3650 -in /tmp/httpschain/cert.csr -CAkey testdata/fake-ca.privkey.pem -CA testdata/fake-ca.cert -passin pass:"gently" -outform pem -out /tmp/httpschain/chain.pem -provider legacy -provider default
+        cat testdata/fake-ca.cert >> /tmp/httpschain/chain.pem
+        cat /tmp/httpschain/chain.pem | jq --raw-input --slurp --compact-output 'split("\n-----END CERTIFICATE-----\n") | map(select(length > 0) | sub("^-----BEGIN CERTIFICATE-----\n"; "") | sub("\n-----END CERTIFICATE-----$"; "")) | { "chain": . }' > /tmp/httpschain/chain.json
```
phbnf, line 156: Wow. Sorry you had to deal with this.

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -116,28 +106,77 @@ resource "google_cloudbuild_trigger" "build_trigger" {
...
+        curl -i -X POST --data @/tmp/httpschain/chain.json -H "Content-Type: application/json" -H "Authorization: Bearer $(cat /workspace/cb_identity)" $(cat /workspace/conformance_url)/ci-${var.project_id}/ct/v1/add-chain
```
phbnf, line 157: Maybe get the output for curl and fail if it's not what expected: a success, and an index of 0.
It's a bit of a rabbit hole, we can go further, wait for a checkpoint, etc. But at this point, I don't think it's worth it - let's just finish building a proper conformance binary.

deployment/modules/gcp/cloudbuild/main.tf
roger2hk: Done.

deployment/modules/gcp/cloudbuild/main.tf
roger2hk: Added more comments.

PR #68, opened 2024-11-28
Enable skipping IAM module for externally managed IAM in GCP Cloud Run

roger2hk:
Towards #53.


roger2hk:
Closing this pull request as the Cloud Run is only used in the Cloud Build which is designed to be used internally in the CI pipeline and has the IAM role mapping defined externally.

PR #84, opened 2025-01-09
Clean serialize.go

phbnf:
Towards #5

This PR cleans serialize.go: dedup related code moves to dedup.go. Since the remaining code is signature related, we can then rename serialize.go to signatures.go, and move


signatures.go
```diff
@@ -20,17 +20,18 @@ import (
...
+	"github.com/google/certificate-transparency-go/x509"
 	"github.com/transparency-dev/formats/log"
-	"github.com/transparency-dev/static-ct/modules/dedup"
-	"golang.org/x/crypto/cryptobyte"
 	"golang.org/x/mod/sumdb/note"
 
 	ct "github.com/google/certificate-transparency-go"
 )
 
+const millisPerNano int64 = 1000 * 1000
```
AlCutter, line 32: Aren't there approximately zero milliseconds per nanosecond, did you mean `nanosPerMilli`? :)

If so, could this be `... = time.Millisecond / time.Nanosecond`?

signatures.go
```diff
@@ -159,62 +160,35 @@ func (cts *CpSigner) KeyHash() uint32 {
...
-				!s.ReadUint16LengthPrefixed(&fingerprints) {
-				return nil, fmt.Errorf("invalid data tile precert_entry")
-			}
-		default:
-			return nil, fmt.Errorf("invalid data tile: unknown type %d", entryType)
-		}
-		k := sha256.Sum256(crt)
-		sctDedupInfo := dedup.SCTDedupInfo{Idx: uint64(i), Timestamp: timestamp}
-		kvs = append(kvs, dedup.LeafDedupInfo{LeafID: k[:], SCTDedupInfo: sctDedupInfo})
+// GetCTLogID takes the key manager for a log and returns the LogID. (see RFC 6962 S3.2)
```
AlCutter, line 186: /key manager/public key/?

signatures.go
```diff
@@ -159,62 +160,35 @@ func (cts *CpSigner) KeyHash() uint32 {
...
+// In CT V1 the log id is a hash of the public key.
+func GetCTLogID(pk crypto.PublicKey) ([sha256.Size]byte, error) {
+	pubBytes, err := x509.MarshalPKIXPublicKey(pk)
+	if err != nil {
+		return [sha256.Size]byte{}, err
```
AlCutter, line 191: nit: You could move the [32]array --> slice conversion from L173 into this method, and change the return type to be `([]byte, error)`, that'd let you return `nil, err` on this line and be a bit clearer.

signatures.go
phbnf: Maybe! This is not new code, I'm blindly migrating it from structures.go. Fixed!

AlCutter: (did you mean to also export it?)

signatures.go
```diff
@@ -20,17 +20,19 @@ import (
...
+	"github.com/google/certificate-transparency-go/x509"
 	"github.com/transparency-dev/formats/log"
-	"github.com/transparency-dev/static-ct/modules/dedup"
-	"golang.org/x/crypto/cryptobyte"
 	"golang.org/x/mod/sumdb/note"
 
 	ct "github.com/google/certificate-transparency-go"
 )
 
+const NanosPerMilli int64 = int64(time.Millisecond / time.Nanosecond)
```
AlCutter, line 33: Did you mean to export it BTW?

phbnf: Woopsie, thanks. As a general comment, the repo needs a good clean-up of exported / unexported stuff as well. I don't want to do this straight away though first I'll re-organize and simplify the code, then I'll carve a few things into an internal/ directory, and then export / unexport accordingly.

AlCutter (approved):

PR #85, opened 2025-01-10
Parse GCP flags in the GCP binary

phbnf:
Towards #5

This PR moves GCP flags parsing to the GCP binary. It also removes the `projectID` flag, which is not used anymore.

### DIFF
--- a/cmd/gcp/main.go
+++ b/cmd/gcp/main.go
@@ -75,9 +75,6 @@ func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
 	ctx := context.Background()
-	if err := validateFlags(); err != nil {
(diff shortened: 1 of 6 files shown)


cmd/gcp/main.go
```diff
@@ -201,16 +201,25 @@ func awaitSignal(doneFn func()) {
 }
 
 func newGCPStorage(ctx context.Context, signer note.Signer) (*sctfe.CTStorage, error) {
+	if len(*bucket) == 0 {
```
AlCutter, line 204: nit: clearer to use `if *bucket == ""`

AlCutter (approved):

PR #91, opened 2025-01-17
Separate chain validation, ct log and http server

phbnf:
Towards #5 

This PR draws boundaries between:
 - chain validation
 - core CT log implementing all the primitives necessary to run a static-ct-api log
 - the HTTP server, implementing static-ct-api HTTP endpoints


ctlog.go
```diff
@@ -61,40 +64,31 @@ type ChainValidationConfig struct {
...
+// TODO(phboneff): consider moving to methods when refactoring logInfo.
+type log struct {
+	// origin identifies the log. It will be used in its checkpoint, and
 	// is also its submission prefix, as per https://c2sp.org/static-ct-api.
-	Origin string
-	// Used to sign the checkpoint and SCTs.
-	Signer crypto.Signer
-	// CertValidationOpts contains various parameters for certificate chain
+	origin string
+	// Signs SCTs.
```
roger2hk, line 76:
```suggestion
	// signSCT signs SCTs.
```

handlers.go
```diff
@@ -207,17 +193,12 @@ type logInfo struct {
 	validationOpts CertValidationOpts
 	// storage stores log data
 	storage Storage
-	// signer signs objects (e.g. STHs, SCTs) for regular logs
-	signer crypto.Signer
+	// signer signs SCTs
+	signSCT signSCT
```
roger2hk, line 197: Do you want to call this `signer` or `signSCT`?

roger2hk (approved):

handlers.go
phbnf: Ah, thanks! Was called signer, and I then renamed it to signSCT.

PR #95, opened 2025-01-22
Simplify LogInfo

phbnf:
Towards #5 

This PR does two main things: 
 - it simplifies `logInfo` by embedding two structs rather than defining all the fields. 
 - it merges the `newLogInfo` into `NewPathHandlers` direclty.


roger2hk (approved):
It's a bit weird to go from info to log. It's more logical if it goes from log and then info.

phbnf:
Yes, I agree with you, a lot of this is here because the CTFE used to support multiple logs, which the SCTFE does not. I'm trying to delete loginfo altogether at the moment, otherwise I'll rename it to something like "instance".

PR #101, opened 2025-02-04
Delete LogInfo object

phbnf:
It contains only two fields, let's inline them.

Towards #5.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -87,8 +87,9 @@ type pathHandlers map[string]appHandler
 // appHandler holds a logInfo and a handler function that uses it, and is
 // an implementation of the http.Handler interface.
 type appHandler struct {
-	info    *logInfo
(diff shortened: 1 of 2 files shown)


handlers.go
```diff
@@ -84,11 +84,12 @@ var entrypoints = []entrypointName{addChainName, addPreChainName, getRootsName}
 // pathHandlers maps from a path to the relevant AppHandler instance.
 type pathHandlers map[string]appHandler
 
-// appHandler holds a logInfo and a handler function that uses it, and is
-// an implementation of the http.Handler interface.
+// appHandler holds a log, common handler options and a handler function.
+// It is an implementation of the http.Handler interface.
```
AlCutter, line 88: Since you're here, perhaps this comment would be improved by updating to talk only about _what_ the struct is/is for - the bits you had to update in this PR is just repeated info from the definition itself below?

AlCutter (approved):

roger2hk (approved):

PR #102, opened 2025-02-05
Move code inside internal directory

phbnf:
This PR moves as many things as possible under /internal.

To achieve this, it also:
 - moves the definition of the `Log` struct `Storage` interface in `handlers.go`, where they are actually used
 - makes a few private fields / structs public, which is ok since they are defined in `/internal`
 - moves `storage.go` to `storage/storage.go`


ctlog.go
```diff
@@ -65,30 +67,25 @@ type ChainValidationConfig struct {
...
+// Now returns the true current local time.
+func (s systemTimeSource) Now() time.Time {
+	return time.Now()
 }
 
-func NewLog(ctx context.Context, origin string, signer crypto.Signer, cfg ChainValidationConfig, ts timeSource, cs CreateStorage) (*log, error) {
-	log := &log{}
+var timeSource = systemTimeSource{}
+
+func NewLog(ctx context.Context, origin string, signer crypto.Signer, cfg ChainValidationConfig, cs CreateStorage) (*scti.Log, error) {
```
AlCutter, line 82: Returning instances of structs from an internal package undermines the primary benefit you get from putting things in there, which is that you're free to refactor to your heart's content because nobody can possibly be relying on the APIs inside. If you "smuggle" instances/funcs out, then you can no longer change that surface as it's now public.

You also cause some weird behaviour for users of the package: e.g. this works:
```go
  myLog := ctlog.New()
  myLog.DoThings()
```
but these won't compile:
```go
struct MyApp {
  myLog *scti.Log // NO
  ...
}
...
     myApp := MyApp{
       myLog: ctlog.New(),
     }
...
```

```go
func main() {
   myLog := ctlog.New()
   doSomething(myLog)
}

func doSomething(l *scti.Log) { // NO
   log("I'm going to do things")
   l.DoThings()
}

Comment (phbnf) [reply ctlog.go:82]: Yeah, I did ponder about it for a long time, and concluded that it was an okay compromise, if only, as a stepping stone. If a struct is defined in the internal directory, it seems reasonable to me to assume that if you're depending on it, you're on your own. The crux of the issue is not really that the return struct is from an internal package, but rather that it has public fields, which could be accessed. If there were all private, it wouldn't be a problem.

Similarly, Tessera public methods export methods which reference private structures in their signature, such as [`WithCheckpointSigner`](https://github.com/transparency-dev/trillian-tessera/blob/main/log.go#L85). The main difference is that these objects can't be used outside of internal because they all take a private object as an argument `options.StorageOptions`, which can't be created outside of the internal package.

I had three solutions in mind before I sent this PR:
1. Move more code from `main.go` into `ctlog.go` such that it registers the HTTP handlers itself. I'm definitely open to it, I thought I wouldn't move too much code in one go and leave this one for later. It's my preferred follow up (I could even send a PR beforehand and rebase everything). The downside of this is that you don't really see what happens under the hood when you look at `main.go`, as opposed to seing clearly that a log is created, then handlers are being created. [Here's what it could look like](https://github.com/phbnf/static-ct/compare/internal...phbnf:static-ct:morehide?expand=1).
2. To make all the fields of this struct private (hence, not accessible), I'd need to move `NewLog` to the internal directory, and then write a wrapper inside `ctlog.go` that calls it. That's what we talked about in London, [here's a quick dirty shot a it](https://github.com/phbnf/static-ct/compare/internal...phbnf:static-ct:internalNoExport?expand=1). I didn't pursue this option after our conversation in London.
3. Leave the Log struct and Storage interface explicitly public, which signs us up to maintaining them. It's my least preferred. 

Do you have any preference?

Comment (AlCutter) [reply ctlog.go:82]: I think (1) is a good plan - returning a `http.Handler` (which internally could be a `http.ServeMux` with all the static CT handlers registered) would let alternative `main` implementations use other stacks (e.g. Gorilla) while being keeping internal stuff private.

I do think that whatever else is returned should also be an exported public thing, this _is_ the public API of the static CT implementation after all (and, otherwise, you'll still have the weirdness of not being able to pass it around, or put it in a declared variable, etc.). But, of course, the public API is not set in stone until we tag a `1.0`, so it's fine to evolve it.

Comment (phbnf) [reply ctlog.go]: Should be better now that #114 has been merged. This likely unblocked further code refactoring, but let's handle them in followup PRs.

File: internal/scti/handlers.go [hunk 2]
```diff
@@ -81,15 +108,15 @@ func setupMetrics(mf monitoring.MetricFactory) {
 // entrypoints is a list of entrypoint names as exposed in statistics/logging.
 var entrypoints = []entrypointName{addChainName, addPreChainName, getRootsName}
 
-// pathHandlers maps from a path to the relevant AppHandler instance.
-type pathHandlers map[string]appHandler
+// PathHandlers maps from a path to the relevant AppHandler instance.
+type PathHandlers map[string]appHandler
```
AlCutter, line 112: Do you still need to export this?

internal/scti/timesource.go
```diff
@@ -12,21 +12,13 @@
...
-package sctfe
+package scti
 
 import "time"
 
-// timeSource can provide the current time, or be replaced by a mock in tests to return
+// TimeSource can provide the current time, or be replaced by a mock in tests to return
 // specific values.
-type timeSource interface {
+type TimeSource interface {
```
AlCutter, line 21: (follow up fine) I think this could just be  `type TimeSource func() time.Time`

ctlog.go
```diff
@@ -66,35 +68,28 @@ type ChainValidationConfig struct {
...
 
-var sysTimeSource = SystemTimeSource{}
+var sysTimeSource = systemTimeSource{}
 
 // newLog instantiates a new log instance, with write endpoints.
 // It initiates chain validation to validate writes, and storage to persist
 // chains.
-func newLog(ctx context.Context, origin string, signer crypto.Signer, cfg ChainValidationConfig, cs CreateStorage) (*log, error) {
-	log := &log{}
+func newLog(ctx context.Context, origin string, signer crypto.Signer, cfg ChainValidationConfig, cs CreateStorage) (*scti.Log, error) {
```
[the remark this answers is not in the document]
phbnf: Moved it there.

ctlog.go
```diff
@@ -125,14 +120,14 @@ func newLog(ctx context.Context, origin string, signer crypto.Signer, cfg ChainV
...
-	log.storage = storage
+	log.Storage = storage
 
 	return log, nil
 }
 
 // newCertValidationOpts checks that a chain validation config is valid,
 // parses it, and loads resources to validate chains.
-func newCertValidationOpts(cfg ChainValidationConfig) (*chainValidationOpts, error) {
+func newCertValidationOpts(cfg ChainValidationConfig) (*scti.ChainValidationOpts, error) {
```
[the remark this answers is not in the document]
phbnf: This one is a bit harder to move: the Config struct has to be defined somewhere that's visible to external libraries.

Eventually, this options are used to call validateChain(chain, options). I think the right thing to do is to re-architecture the code for it to call methods (see [TODO](https://github.com/transparency-dev/static-ct/blob/4331ec4a589eced34f2a828c3136786a616a2a26/ctlog.go#L72)). I'll come to this later. This is specifically cumbersome to do for this one because these options are also used for the getRoots endpoint.... which is already a tricky one to handle with the x509 migration. So one step at a time :)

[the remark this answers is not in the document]
phbnf: What I've done though, is to move the OID and ExtKeyUsage parsing functions to chain_validation.go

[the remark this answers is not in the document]
phbnf: In the meantime, to make this PR easier to digest, I can add a constructor in internal. That will save the PR from all the lowercase-->uppercase changes.

AlCutter (approved):

Issue #105, opened 2025-02-06
Export monitoring metrics


phbnf:
- [x] Review the current metrics and modify them accordingly
- [x] Add new metrics relevant to Tessera
- [x] Take a dep on prometheus directly rather than using an interface
- [x] migrate to Open Telemetry
- [x] Add deduplication breakdown
- [ ] #521

phbnf:
We eventually migrated to Open Telemetry metrics. We need to add some metrics for chain storage.

phbnf:
Adding issuer storage metrics required to re-work a few things in the code, I filed #521.

PR #117, opened 2025-02-10
Drop dependency on trillian/monitoring

phbnf:
Towards #105 and #106

This PR makes use of the prometheus library directly, rather than using trillian/monitoring.

It also introduces the following changes:
 - a new `lastSCTIndex` metric since SCTs now include an index
 - changes the `logid` label to `origin`
 - changes the label `ep` label to `op`

### DIFF
--- a/internal/scti/handlers.go
+++ b/internal/scti/handlers.go
@@ -30,7 +30,8 @@ import (
 
 	"github.com/google/certificate-transparency-go/tls"
 	"github.com/google/certificate-transparency-go/x509"
-	"github.com/google/trillian/monitoring"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/transparency-dev/static-ct/modules/dedup"
 	tessera "github.com/transparency-dev/trillian-tessera"
 	"github.com/transparency-dev/trillian-tessera/ctonly"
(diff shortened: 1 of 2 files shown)


internal/scti/handlers.go
```diff
@@ -62,20 +63,53 @@ var (
 	// Metrics are all per-log (label "origin"), but may also be
 	// per-entrypoint (label "ep") or per-return-code (label "rc").
 	once             sync.Once
-	knownLogs        monitoring.Gauge     // origin => value (always 1.0)
-	lastSCTTimestamp monitoring.Gauge     // origin => value
-	reqsCounter      monitoring.Counter   // origin, ep => value
-	rspsCounter      monitoring.Counter   // origin, ep, rc => value
-	rspLatency       monitoring.Histogram // origin, ep, rc => value
+	knownLogs        *prometheus.GaugeVec     // origin => value (always 1.0)
```
AlCutter, line 66: Could just remove this one, I suspect?

AlCutter (approved):

internal/scti/handlers.go
phbnf: I've left it on purpose: this will allow to build global dashboard with all the log names for instance, if multiple stacks exist. It's a nice way to extract the origin, and protects us from bugs like nan values on some metrics because no request is flying around.

Issue #59, opened 2024-11-21
Conformance + Load testing infrastructure


phbnf:
We need to both have a conformance and load test for the SCTFE binary.

Today, we have:
  - the [certificate-transparency-go/.../ct_hammer](https://github.com/google/certificate-transparency-go/blob/master/trillian/integration/ct_hammer/main.go) tool. It allows to send generate and send `add-chain` and `add-pre-chain` requests, which work with the SCTFE with no modification. However, the read paths are not compatible, so it does not work as a conformance tool.
  - the [trillian-tessera/.../hammer](https://github.com/transparency-dev/trillian-tessera/blob/main/internal/hammer/hammer.go) tool, which we use for conformance testing and load testing Tessera. To works with CT, it needs to:
    1. support different `add` endpoints: `add-chain` and `add-prechain` instead
    2. support different read endpoints for leaves, which are stored under `/tile/data` instead of `/tile/entries`
    3. parse leaf bundles in a different way, since they are sequences of `TileLeaf`, and instead of sequences of `big-endian uint16 length-prefixed` entries
    4. support static-ct api checkpoints, instead of classic checkpoints
    5. support generating certificates chains instead of generating regular entries. Note that these chains need to also be compatible with the nice deduplication tests that the hammer supports. That might be a bit of a challenge for CT given that certs generation might not be idempotent for various reasons.

phbnf:
A few weeks ago, I wrote two experiment PRs, that handled i, ii, iii, iv and v.

Changes for i, ii, iii, and iv are not that big, but may require further rearchitecturing shall we decide for the hammer tool to optionally supports CT with flags for instance.

I did something a bit special for v. that I think is out of scope of this issue (though it might be a nice-to-have): instead of generating chains locally, it plugged the [certificate-transparency-go/.../scanlog](https://github.com/phbnf/certificate-transparency-go/blob/master/scanner/scanlog/scanlog.go) tool into [trillian-tessera/.../hammer](https://github.com/transparency-dev/trillian-tessera/blob/main/internal/hammer/hammer.go). Since a lot of things have changed on the hammer since I wrote this, and especially that would impact point v. with respect to deduplication, I think that it won't work anymore. But I'll give it a go when I come around to it. For history and inspiration, here are the two PRs that resulted from this, which I have rebased on main/master. 
[...]

PR #129, opened 2025-02-12
Add CT Hammer testdata generation and files

roger2hk:
Towards #59.

This testdata gen tool generates the following test files:
- Root CA RSA private key
- Root CA certificate
- Intermediate CA RSA private key
- Intermediate CA certificate
- Leaf certificate signing RSA private key


internal/hammer/testdata/test_root_ca_private_key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
phbnf, line 1: The Go x509 library uses "TESTING KEY" instead of `PRIVATE KEY`, and they uses a small function to replace this in the code. Could we use this? I quite like that! Example:  https://cs.opensource.google/go/go/+/master:src/crypto/x509/pem_decrypt_test.go;l=83?q=TESTING%20KEY&ss=go%2Fgo:src%2Fcrypto%2Fx509%2F

internal/hammer/testdata/gen.go
```diff
@@ -0,0 +1,184 @@
...
+	if err := saveRSAPrivateKeyPEM(leafCertPrivateKey, path.Join(*outputPath, "test_leaf_cert_signing_private_key.pem")); err != nil {
+		klog.Fatalf("Failed to save leaf certificate signing private key: %v", err)
+	}
+}
+
+func rootCACert(privKey *rsa.PrivateKey) (*x509.Certificate, error) {
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			Organization: []string{fmt.Sprintf("%s Root CA", organization)},
```
phbnf, line 98: Let's add "test" somewhere? Maybe in the var at the top of the file?

internal/hammer/testdata/gen.go
```diff
@@ -0,0 +1,184 @@
...
+		klog.Fatalf("Failed to save intermediate CA certificate: %v", err)
+	}
+
+	// Generate a new RSA leaf certificate signing private key.
+	leafCertPrivateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		klog.Fatalf("Failed to generate leaf certificate signing private key: %v", err)
+	}
+	if err := saveRSAPrivateKeyPEM(leafCertPrivateKey, path.Join(*outputPath, "test_leaf_cert_signing_private_key.pem")); err != nil {
+		klog.Fatalf("Failed to save leaf certificate signing private key: %v", err)
```
phbnf, line 90: What's that going to be used for? Should we re-generate it when new leaf certs are being generated?

internal/hammer/testdata/gen.go
roger2hk: Added `Test` to root CA certificate and intermediate CA certificate generation.

internal/hammer/testdata/test_root_ca_private_key.pem
roger2hk: Sure. I've added `TEST` in between.

roger2hk: The small function will be added in CT hammer binary.

phbnf: I don't know for sure, but I can see how automatic scanners would alert on a string match on "PRIVATE KEY". Let's see how we get on, we can always change it later.

internal/hammer/testdata/gen.go
phbnf: Let's remove this for now unless there's a specific reason why we need it to be stable across runs.

roger2hk: Removed.

phbnf (approved):

PR #139, opened 2025-02-17
Copy client code from transparency-dev/trillian-tessera

roger2hk:
Towards https://github.com/transparency-dev/static-ct/issues/59.

This is a direct copy of client from transparency-dev/trillian-tessera. This should help the ongoing pull request review easier by highlighting the differences later.


phbnf:
Is it expected that the test fails? Is there a quick way to fix them in a follow-up PR, or even in a separate commit in this PR?

roger2hk:
> Is it expected that the test fails? Is there a quick way to fix them in a follow-up PR, or even in a separate commit in this PR?

Oh, I forgot to copy the test data. The test data will be updated to align with the Static CT API specification when the client logic is updated later.

phbnf (approved):

PR #150, opened 2025-02-24
Regenerate test certificates to get rid of RSA SHA1

phbnf:
Towards #120.

New certs use ECDSA. Lots of credits go to @roger2hk for https://github.com/transparency-dev/static-ct/tree/main/internal/hammer/testdata.

This will allow to migrate overt c-t-go/x509 to crypto/x509.

I also had to slightly modify the test in `chain_validation_test.go`.

### DIFF
--- a/internal/testdata/gen/gen.go
+++ b/internal/testdata/gen/gen.go
@@ -47,9 +47,9 @@ var (
 
 var (
 	// From RFC6962 Section 3.1. To identify pre-certs.
-	cTPrecertPoisonOID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 3}
+	ctPrecertPoisonOID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 3}
 	ctPoison           = pkix.Extension{
-		Id:       cTPrecertPoisonOID,
+		Id:       ctPrecertPoisonOID,
 		Critical: true,
 		Value:    []byte{0x05, 0x00}, // ASN.1 NULL
 	}
@@ -273,7 +273,7 @@ func (g *chainGenerator) certificate(serialNumber int64, preCert bool, notBefore
 	}
 
 	ctPoison := pkix.Extension{
-		Id:       cTPrecertPoisonOID,
+		Id:       ctPrecertPoisonOID,
 		Critical: true,
 		Value:    []byte{0x05, 0x00}, // ASN.1 NULL
 	}


internal/testdata/gen/gen.go
```diff
@@ -0,0 +1,393 @@
...
+	// Fix block type for testing keys.
+	block.Type = testingKey(block.Type)
+
+	switch block.Type {
+	case "EC PRIVATE KEY":
+		return x509.ParseECPrivateKey(block.Bytes)
+	default:
+		return nil, fmt.Errorf("unsupported PEM block type: %s", block.Type)
+	}
+}
```
AlCutter, line 349: Is this used?

internal/testdata/gen/gen.go
```diff
@@ -0,0 +1,393 @@
...
+		klog.Info("Warning: More than one PEM block found. Parsing only the first.")
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse X.509 certificate: %w", err)
+	}
+
+	return cert, nil
+}
```
AlCutter, line 374: Is this used?

internal/testdata/gen/gen.go
```diff
@@ -0,0 +1,393 @@
...
+		return k.Public()
+	case *ecdsa.PrivateKey:
+		return k.Public()
+	case *ed25519.PrivateKey:
+		return k.Public()
+	default:
+		klog.Fatalf("Unknown private key type: %T", privKey)
+		return nil // Or panic, or return an error
+	}
+}
```
AlCutter, line 389: Used?

internal/testdata/gen/gen.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	return cert, nil
+}
+
+func saveECDSAPrivateKeyPEM(key *ecdsa.PrivateKey, filename string) error {
+	// Marshal the private key to SEC1 ASN.1 DER.
+	derBytes, err := x509.MarshalECPrivateKey(key)
+	if err != nil {
+		return fmt.Errorf("failed to marshal EC private key: %w", err)
```
AlCutter, line 291: `%w` -> `%v` unless you need to have callers be able to distinguish between different error instances/types (throughout)

internal/testdata/gen/gen.go
```diff
@@ -0,0 +1,393 @@
...
+	pemData := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: cert.Raw,
+	})
+
+	if err := os.WriteFile(filename, pemData, 0644); err != nil {
+		return fmt.Errorf("failed to write PEM file: %w", err)
+	}
+	// TODO(phboneff): automate cert printin in certificate.go
+	fmt.Println("Don't forget to update certificate.go with the matching openssl text output.")
```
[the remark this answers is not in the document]
phbnf: Yes, I gave it a try, and it works. I'm not doing this in this CL since:
 - It's not super practical, ideally, I'd like to have it in certificates.go to be able to scroll between all the certs that make up a chain.
 - [certificate-transparency-go/x509util
/x509util.go](https://github.com/google/certificate-transparency-go/blob/d684d515953e5a2ce95620b9955068657e75ab19/x509util/x509util.go#L403)
has printing functions. I did not move them over as it involves a lot of code. We could leave a dep on c-t-go just for this, or use other libraries like https://github.com/smallstep/certinfo, https://github.com/grantae/certinfo, https://github.com/pete911/certinfo or any better one you're aware of. I didn't want to mix this with this PR.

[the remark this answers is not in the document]
AlCutter: Fair enough to not do it in this PR, but I'd be cautious of leaving it in the `.go` file as it's almost certainly going to rot if we ever re-generate the certs again.

Anyway, not a blocker, mostly just capturing the thought.

internal/testdata/gen/gen.go
```diff
@@ -0,0 +1,393 @@
...
+		intermediateCert:  intermediateCert,
+		intermediateKey:   intermediateKey,
+		leafCertPublicKey: leafCertPublicKey,
+	}
+}
+
+// certificate generates a deterministic TLS certificate by using integer as the serial number.
+func (g *chainGenerator) certificate(serialNumber int64, preCert bool) (*x509.Certificate, error) {
+	notBefore := time.Date(2023, 1, 1, 0, 0, 0, 0, time.UTC)
+	notAfter := time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC)
```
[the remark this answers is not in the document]
phbnf: Ah yes reminded me I forgot to get back to this one. Fixed now. It's ok for now given that the notAfter/notBefore tests in the SCTFE use other certificates, so it did not impact this. I've removed the setting to reject expired certs ont the SCTFE since we don't use this, notAfter / notBefore limits are enough.

One thing that was very odd though and that I had forgotten, is that root notBefore/notAfter did not match with the leaf certs. That's ok for CT though since it doesn't apply time checks when it validates the chain...

To avoid any further confusion, I have:
 - added a flag to control the notBefore start date with a sensible stable default value (👀 ) 
 - aligned dates for all the certs in the chain

internal/testdata/gen/gen.go
```diff
@@ -43,12 +41,29 @@ const (
...
 
 var (
-	outputPath = flag.String("output_path", "./internal/testdata/", "Output path for private keys and certificates")
+	outputPath      = flag.String("output_path", "./internal/testdata/", "Output path for private keys and certificates")
+	notBeforeString = flag.String("not_before", "2024-12-05T18:05:50.000Z", "Start of the range of certs to be generated. RFC3339 UTC format, e.g: 2024-01-02T15:04:05Z.")
+)
+
+var (
+	// From RFC6962 Section 3.1. To identify pre-certs.
+	cTPrecertPoisonOID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 3}
```
AlCutter, line 50: nit: `ctPrecert...` to match the `ctPoison` case-style below

AlCutter (approved):
Just one tiny nit!

internal/testdata/gen/gen.go
phbnf: good catch, thanks!

PR #157, opened 2025-02-25
X509lax

phbnf:
Toward #120.

This PR copies the smallest amount of code we can, allowing to build Tesseract without a dep on c-t-go/x509, and using "crypto/x509" wherever it can.

It is WIP, but ready for someone to take a first pass, and confirm that this PR can easily be reviewed.
A followup PR will migrate the codebase to this fork + "crypto/x509". You can have a sneak peek here: https://github.com/phbnf/static-ct/tree/evenmoreold. 

This PR has two commits. To review this PR, review commits independtly.
 - one that copies `verify.go` and `cert_pool.go` from "crypto/x509" Not much to review there, it's a copy.
 - one that removes all the things we either don't need to run a CT log, or would break a CT log. Here are all the modification it brings:
      - Deletes checks that would not pass with certificates submitted to CT logs, such as criticalExtensions checks, because pre-certs have a critical pre cert extension that the crypto/x509 library does not recognize
       - changes package name
       - takes a dep on "crypto/x509" whenever we can
      - modify some method signatures from `(c *Certificate) func(args)` to `func(c x509.Certificate, args)` to re-use the `x509.Certificate` object without redefining it
      - sets struct field names in declaration when not set such as: `CertificateInvalidError{c} --> CertificateInvalidError{Cert: c}`
      - copies over a very small oid definition from `x509.go`. I've left it in a small `x509.go` file just to avoid extra modifications.
      
      
TODO(phboneff):
 - [x] update to go 1.24, and use NoValidChains
 - [x]  remove some TODOs
 - [x] add Readme that explains what this package is, and why it's there

### DIFF
--- a/internal/x509fork/README.md
+++ b/internal/x509fork/README.md
@@ -2,9 +2,11 @@
 
 This is a minimalist fork of [`crypto/x509`](https://pkg.go.dev/crypto/x509). 
 
-[!WARNING] This library is not safe to use for applications outside of this repository.
+> [!WARNING]
+> This library is not safe to use for applications outside of this repository.
 
-[!WARNING] This fork will not be kept in synced with upstream. It will not be updated, unless required by a security vulnerability or a critical functionality issue.
+> [!WARNING]
+> This fork will not be kept in synced with upstream. It will not be updated, unless required by a security vulnerability or a critical functionality issue.
 
 As specified by [RFC6962 S3.1](https://www.rfc-editor.org/rfc/rfc6962#section-3.1), CT logs MUST validate sumbmitted chains to ensure that they link up to roots they accept. `crypto/x509` implements this, and also runs additional common chain validation checks. However, these additional checks:
  - Do not allow chains to contain precertificates or preissuers intermediates.
--- a/internal/x509fork/README.md
+++ b/internal/x509fork/README.md
@@ -1,0 +1,20 @@
+# x509fork
+
+This is a minimalist fork of [`crypto/x509`](https://pkg.go.dev/crypto/x509). 
+
+[!WARNING] This library is not safe to use for applications outside of this repository.
+
+[!WARNING] This fork will not be kept in synced with upstream. It will not be updated, unless required by a security vulnerability or a critical functionality issue.
+
+As specified by [RFC6962 S3.1](https://www.rfc-editor.org/rfc/rfc6962#section-3.1), CT logs MUST validate sumbmitted chains to ensure that they link up to roots they accept. `crypto/x509` implements this, and also runs additional common chain validation checks. However, these additional checks:
+ - Do not allow chains to contain precertificates or preissuers intermediates.
+ - Would block non compliant certificates signed by production roots from being accepted, therefore preventing them from becoming transparency discoverable.
+
+The fork in this directory implements chain verification requirements from [RFC6962 S3.1](https://www.rfc-editor.org/rfc/rfc6962#section-3.1) and disables some additional check, such as:
+
+  - **Handling of critical extensions**: CT precertificates are identified by a critical extension defined in [RFC6962 S3.1](https://www.rfc-editor.org/rfc/rfc6962#section-3.1), which the `crypto/x509` library does not process. A non-processed critical extension would fail certificate validation. This check is disabled to allow precertificate in the logs.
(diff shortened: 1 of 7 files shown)


internal/x509util/verify.go
```diff
@@ -240,476 +66,37 @@ const (
...
-				return nil
-			})
-
-			if err != nil {
-				return err
-			}
-		}
-	}
+	// CANotAuthorizedForThisName check deleted.
+	// Allow to log all certificates, even if they have been isued by a CA that
```
AlCutter, line 96: nit: "Allow logging of all ..."

AlCutter, line 96: isued/issued

internal/x509util/verify.go
```diff
@@ -240,476 +66,37 @@ const (
...
+	// is not auhotized to issue certs for a given domain.
```
AlCutter, line 97: /auhotized/authorized/ :)

internal/x509util/x509.go
```diff
@@ -0,0 +1,5 @@
+package x509util
+
+var (
+	oidExtensionSubjectAltName = []int{2, 5, 29, 17}
```
AlCutter, line 4: Is this going to be used?

AlCutter (approved):
This is really cool, I like the direction - _huge_ thanks for putting the substantial effort in here to drop the dep on the old fork!

We should probably be explicit in the pkg doc about the intention here for future folks coming along - e.g.
 * not a full validating x509 impl for TLS use, but bare bones for running a CT log; should not be reused elsewhere for safety reasons.
 * Not likely to be frequently/if at all updated from the upstream stdlib x509
etc.

roger2hk (approved):

internal/x509util/verify.go
roger2hk: I thought the `verify.go` was deleted in the next commit.

internal/x509util/x509.go
phbnf: Yes, it's currently used by `verify.go`.

I could put a description indeed, but it would literally be the variable name..."oid for Subject Alternative Name extension"
 
Note that the [crypto/x509](https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/crypto/x509/x509.go;l=1110) library doesn't have a description, neither does our [c-t-go fork](https://github.com/google/certificate-transparency/blob/ebe4199bb851bcae1fe824ffd8745bb4dc70ede4/go/x509util/x509util.go#L74).

It's the only thing we need from `x509.go`.. I could put it in a different file, I've decided to leave it there to stay as close as possible to upstream.

internal/x509util/verify.go
phbnf: The next PR will start to use `verify.go`, rather than using `verify.go` from the `crypto/x509` library. Otherwise, the `crypto/x509` library would not verify chains submitted to CT logs.

phbnf:
@roger2hk, @AlCutter PTAL.

I have:
 - addressed comments
 - moved the files to their own x509fork package to keep this fork well separated from the rest of the codebase
 - made very minor code changes
 
 I also have a [followup PR ready](https://github.com/phbnf/static-ct/compare/x509lax...phbnf:static-ct:x509laxtest?expand=1) moving tests over. I've split it to make the reviews easier, but tests do pass.

internal/x509fork/README.md
```diff
@@ -0,0 +1,20 @@
+# x509fork
+
+This is a minimalist fork of [`crypto/x509`](https://pkg.go.dev/crypto/x509). 
+
+[!WARNING] This library is not safe to use for applications outside of this repository.
```
roger2hk, line 5:
```suggestion
> [!WARNING] 
> This library is not safe to use for applications outside of this repository.
```

internal/x509fork/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+[!WARNING] This fork will not be kept in synced with upstream. It will not be updated, unless required by a security vulnerability or a critical functionality issue.
```
roger2hk, line 7:
```suggestion
> [!WARNING] 
> This fork will not be kept in synced with upstream. It will not be updated, unless required by a security vulnerability or a critical functionality issue.
```

roger2hk (approved):

AlCutter (approved):

PR #167, opened 2025-03-07
Add MMD verification in CT Hammer

roger2hk:
Towards #112, #113.


internal/hammer/loadtest/client.go
```diff
@@ -155,42 +155,42 @@ type httpLeafWriter struct {
 	bearerToken string
 }
 
-func (w httpLeafWriter) Write(ctx context.Context, newLeaf []byte) (uint64, error) {
+func (w httpLeafWriter) Write(ctx context.Context, newLeaf []byte) (uint64, uint64, error) {
```
mhutchinson, line 158: If the end goal is that this `loadtest` library is to be reused between at least Tessera and TesseraCT then this return type will need to be revised here. But we can merge this now and do the diffs when this is feature complete to see what we need to do.

internal/hammer/loadtest/hammer.go
```diff
@@ -31,27 +31,35 @@ type HammerOpts struct {
...
 	})
 	fullReaders := NewWorkerPool(func() Worker {
 		return NewLeafReader(tracker, f, MonotonicallyIncreasingNextLeaf(), readThrottle.TokenChan, errChan)
 	})
 	writers := NewWorkerPool(func() Worker {
-		return NewLogWriter(w, gen, writeThrottle.TokenChan, errChan, seqLeafChan)
+		return NewLogWriter(w, gen, writeThrottle.TokenChan, errChan, seqLeafChan, leafMMDChan)
+	})
+	mmdVerifiers := NewWorkerPool(func() Worker {
+		return NewMMDVerifier(tracker, errChan, leafMMDChan, opts.MMDDuration)
```
mhutchinson, line 54: nit: looking at this signature compared to the others, it stands out because the others have their channels passed in last. I'm not going to die on this hill, but it would be good if they were somewhat consistent.

internal/hammer/loadtest/workers.go
```diff
@@ -208,18 +229,39 @@ func (w *LogWriter) Run(ctx context.Context) {
...
+		for _, cert := range req.Chain {
+			chain = append(chain, cert...)
+		}
+
+		// Send LeafMMD for inclusion proof verification.
+		select {
+		case w.leafMMDChan <- LeafMMD{chain, index, timestamp}:
+		default:
+			// Drop if leafMMDChan is full. This could happen if the number
+			// of MMD verifiers is smaller than the number of writers.
```
mhutchinson, line 262: nit: More precisely, it happens if the MMD verifiers are falling behind. The numbers are one factor, but the cost of the work is as important. i.e. this could fail even if you had twice as many verifiers as writers, if the cost of verifying was 3x the cost of writing.

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +273,178 @@ func (w *LogWriter) Kill() {
...
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+
+		var leafMMD *LeafMMD
+		for {
+			if leafMMD == nil {
+				mmd, ok := <-v.leafMMDChan
```
mhutchinson, line 313: There's nothing to allow this line to exit if the context is cancelled or the verifier is `Kill`ed. I don't see how this two argument receive operator will ever return false, as the channel isn't closed anywhere? https://go.dev/ref/spec#Receive_operator

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +273,178 @@ func (w *LogWriter) Kill() {
...
+				leafMMD = &mmd
+			}
+
+			// Retry if the leaf is not yet integrated into the log.
+			if leafMMD.index >= v.tracker.LatestConsistent.Size {
+				// Verify MMD timestamp.
+				sec := int64(leafMMD.timestamp / 1000)
+				nsec := int64((leafMMD.timestamp % 1000) * 1000000)
+				if time.Unix(sec, nsec).Add(v.mmdDuration).Before(time.Now()) {
+					v.errChan <- fmt.Errorf("leaf index %d MMD violation at %d", leafMMD.index, leafMMD.timestamp)
```
mhutchinson, line 326: Does this also need to nil out `leafMMD` ? I think this will just get stuck in a loop forever on this entry now?

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +273,178 @@ func (w *LogWriter) Kill() {
...
+				}
+
+				break
+			}
+
+			// A new proof builder is required because v.tracker.ProofBuilder
+			// is not thread safe.
```
mhutchinson, line 333: This comment doesn't make much sense in this context?

The proof builder is constructed with a checkpoint, which is super important. As long as the proof builder was only ever used in this worker, you _could_ keep this proof builder about and only invalidate it when the checkpoint from the tracker needs to be bumped. This would be more efficient and if you had a run of entries in a row that could all be proved to be included in the last used checkpoint, then you could avoid some work. However, the state management would be a little more delicate.

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +273,178 @@ func (w *LogWriter) Kill() {
...
+			size := v.tracker.LatestConsistent.Size
+			hash := v.tracker.LatestConsistent.Hash
```
mhutchinson, line 335: I strongly encourage reading `v.tracker.LatestConsistent` to a variable early on, and then using it through a single run through the loops here. As it stands, the tracker could update between these two calls and you'd save the old size and the new hash, which would be totally broken.

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +273,178 @@ func (w *LogWriter) Kill() {
...
+			pb, err := client.NewProofBuilder(ctx, log.Checkpoint{
+				Origin: v.tracker.Origin,
+				Size:   size,
+				Hash:   hash,
+			}, v.tracker.TileFetcher)
+			if err != nil {
+				v.errChan <- fmt.Errorf("failed to create proof builder: %w", err)
```
mhutchinson, line 342: Again with all of these errors, you need to consider whether `leafMMD` should be nilled out, or whether you think it might pass on subsequent attempts. As written, this looks like it could get stuck in a death loop given a bad entry. Maybe that's a feature, maybe that's a bug.

mhutchinson (commented):
I'm happy with the shape of this, so nicely done. There are a few issues around parallelism that give me some concern though, so PTAL at my comments. We can discuss 1:1 if you want more details.

internal/hammer/loadtest/workers.go
phbnf: Yes - and we might want to add a logging message here to highlight that some entries are not sent for analysis.

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +273,178 @@ func (w *LogWriter) Kill() {
...
+// Kills this verifier at the next opportune moment.
+// This function may return before the verifier is dead.
+func (v *MMDVerifier) Kill() {
+	if v.cancel != nil {
+		v.cancel()
+	}
+}
+
+// entryFromChain generates an Entry from a chain and timestamp.
+// copied from certificate-transparency-go/serialization.go
```
phbnf, line 381: Just to keep track of things, can you add a comment / TODO to reflect that it comes form handlers.go?

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +273,178 @@ func (w *LogWriter) Kill() {
...
+				}
+				leafMMD = &mmd
+			}
+
+			// Retry if the leaf is not yet integrated into the log.
+			if leafMMD.index >= v.tracker.LatestConsistent.Size {
+				// Verify MMD timestamp.
+				sec := int64(leafMMD.timestamp / 1000)
+				nsec := int64((leafMMD.timestamp % 1000) * 1000000)
+				if time.Unix(sec, nsec).Add(v.mmdDuration).Before(time.Now()) {
```
phbnf, line 325: I think you can use `time.UnixMilli` directly with `leafMMD.timestamp`.

internal/hammer/loadtest/client.go
roger2hk: Correct, the end goal is to use the `loadtest` library from trillian-tessera repo.

internal/hammer/loadtest/hammer.go
roger2hk: It's good to maintain the consistency. Rearranged the `NewMMDVerifier` arguments.

internal/hammer/loadtest/workers.go
roger2hk: Good catch. It will not stuck since there is a `break` after this.

However, it looks like that `break` will stop the `retry` of the same `leafMMD`. I've updated the `break` and `continue`. Please take another look.

The `break` now should be equivalent to set `leafMMD` to `nil`.

internal/hammer/loadtest/workers.go
roger2hk: Good catch. Updated to use `v.tracker.LatestConsistent`.

internal/hammer/loadtest/workers.go
roger2hk: Updated. I'm not sure if we shoud `panic` or fail with fatal log. It is expected to always build the `ProofBuilder` successfully.

internal/hammer/loadtest/workers.go
roger2hk: Added the comment.

internal/hammer/loadtest/workers.go
roger2hk: Good call, updated.

internal/hammer/loadtest/workers.go
roger2hk: Right, I hit into the checkpoint change issue. That's why I create a new proof builder with the last checkpoint here. As the performance is doing alright at the moment, I've added a TODO.

internal/hammer/loadtest/workers.go
roger2hk: Good catch. This part was referenced from [analysis.go#L75-L78](https://github.com/transparency-dev/static-ct/blob/5fbf0e90d516f75ee866a0b2c81412b669cf4a4d/internal/hammer/loadtest/analysis.go#L75-L78). Should I remove it or keep it just in case this will be closed somewhere later?

mhutchinson: My main concern is that if you started the hammer with a huge number of these verifiers and then scaled them down, all of them would be blocked on this line before any entries are written to the `leafMMDChan`. When scaling them down (i.e. calling `Kill` on them), they would still be blocked on this line waiting for an entry to become available, or the channel to be closed.

When writing background workers, the potential blocking operations should all respect cancel signals as soon as possible. One way to do this could be to wrap this in a select block and add a case for `ctx.Done` too. But as there's already a similar select/case just outside the loop I expect there's a way to merge these together. All the below is speculative but I can look at this closer if needed:

Can we get rid of the inner for loop, and have code with the following shape?:

```
ctx, v.cancel = context.WithCancel(ctx)
var leafMMD *LeafMMD // the leaf we are currently processing, if any
for {
  if leafMMD == nil {
    select {
      case <-ctx.Done():
        return
      case mmd, ok := <-v.leafMMDChan:
        if !ok {
          break
        }
        leafMMD = &mmd
    }
  } else {
    // We have a leaf but failed to find it integrated last time, wait a bit and try again
    time.Sleep(500ms)
  }
  // invariant: leafMMD is not nil, and now we are going to see if it's integrated....
}

File: internal/hammer/loadtest/hammer.go [hunk 11]
```diff
@@ -31,27 +31,35 @@ type HammerOpts struct {
...
 	NumWriters       int
+	NumMMDVerifiers  int
+	MMDDuration      time.Duration
 }
 
 func NewHammer(tracker *client.LogStateTracker, f client.EntryBundleFetcherFunc, w LeafWriter, gen func() []byte, seqLeafChan chan<- LeafTime, errChan chan<- error, opts HammerOpts) *Hammer {
 	readThrottle := NewThrottle(opts.MaxReadOpsPerSecond)
 	writeThrottle := NewThrottle(opts.MaxWriteOpsPerSecond)
 
+	leafMMDChan := make(chan LeafMMD, opts.NumWriters)
```
phbnf, line 42: Should this be higher? It looks to me that this channel will get full, new `leafMMD` won't be added to it, and this behavior would go silent. As we spoke about, logging a message each time a `leafMMD` would spam logs when `MMDVerifiers` are not enabled. I think that would be fine provided that the hammer should always have them on. Alternatively, you could compare the length and capacity of the channel at read time to tell whether or not it's full, and indicate to the operator that they might want to increase the channel size?

If there aren't enough `MMDVerifier` workers (which could be the case because integration is slower than sequencing) and the channel gets full, you might end up always verifying MMDs after MMD+X time has passed, and therefore never detect if there is an MMD violation smaller between [0;X]

internal/hammer/loadtest/workers.go
roger2hk: Thanks for the suggestion. I've updated the code.

internal/hammer/loadtest/workers.go
```diff
@@ -258,8 +258,7 @@ func (w *LogWriter) Run(ctx context.Context) {
 		select {
 		case w.leafMMDChan <- LeafMMD{chain, index, timestamp}:
 		default:
-			// Drop if leafMMDChan is full. This could happen if the number
-			// of MMD verifiers is smaller than the number of writers.
+			// Drop if leafMMDChan is full. This could happen if the MMD verifiers are falling behind.
```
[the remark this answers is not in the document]
roger2hk: Added the log when the `leafMMDChan` is full.

internal/hammer/loadtest/hammer.go
roger2hk: Bumped the `leafMMDChan` buffer size to a huge size. I've stop logging when the number of MMD verifiers is zero.

internal/hammer/loadtest/workers.go
```diff
@@ -208,18 +229,41 @@ func (w *LogWriter) Run(ctx context.Context) {
...
+		for _, cert := range req.Chain {
+			chain = append(chain, cert...)
+		}
+
+		// Send LeafMMD for inclusion proof verification.
+		select {
+		case w.leafMMDChan <- LeafMMD{chain, index, timestamp}:
+		default:
+			// Drop if leafMMDChan is full. This could happen if the MMD verifiers are falling behind.
+			if len(w.leafMMDChan) > 0 {
```
mhutchinson, line 262:
```suggestion
			if cap(w.leafMMDChan) > 0 {
```

The length of a channel can fluctuate. I think you're after the assigned capacity?

internal/hammer/loadtest/hammer.go
```diff
@@ -31,27 +31,35 @@ type HammerOpts struct {
...
 	NumWriters       int
+	NumMMDVerifiers  int
+	MMDDuration      time.Duration
 }
 
 func NewHammer(tracker *client.LogStateTracker, f client.EntryBundleFetcherFunc, w LeafWriter, gen func() []byte, seqLeafChan chan<- LeafTime, errChan chan<- error, opts HammerOpts) *Hammer {
 	readThrottle := NewThrottle(opts.MaxReadOpsPerSecond)
 	writeThrottle := NewThrottle(opts.MaxWriteOpsPerSecond)
 
+	leafMMDChan := make(chan LeafMMD, opts.NumWriters*opts.NumMMDVerifiers)
```
mhutchinson, line 42: This is a unique way of picking the capacity of the channel. The good news is that it sets the capacity to 0 if there are no verifiers, but I don't like the implication that we're trying to create a channel big enough so that every writer can send a message to every verifier.

internal/hammer/loadtest/workers.go
roger2hk: Good catch! Updated.

internal/hammer/loadtest/hammer.go
roger2hk: I've reverted it to use the number of writers. If the number of MMD verifiers is 0, the `leafMMDChan` is `nil` instead of buffer size `0`. The `cap(leafMMDChan)` returns 0 when `leafMMDChan = nil`.

roger2hk: I've tracked the number of entries in the channel. It does not reach the number of writers * 12 when there are enough MMD verifiers to process the channel.

internal/hammer/loadtest/workers.go
```diff
@@ -231,3 +275,187 @@ func (w *LogWriter) Kill() {
...
+			select {
+			case <-ctx.Done():
+				return
+			case mmd, ok := <-v.leafMMDChan:
+				if !ok {
+					return
+				}
+				leafMMD = &mmd
+			default:
+				// Nothing in the leafMMDChan, wait a bit and try again.
```
mhutchinson, line 318: The purpose of selecting from the two channels above is that you peacefully block until cancellation or availability of work. This default case causes this to be busier by running around the loop more often... why?

roger2hk: Ah right, I've removed it.

mhutchinson (approved):

phbnf (approved):

Issue #103, opened 2025-02-06
Improve test coverage


phbnf:

phbnf:
After review, we can add tests to:
 1. tesseract/storage
 2. `signatures.go`, and `handlers.go` with a broken backend
 3. Eventually, add tests to x509.go

phbnf:
For reference:

```
└─> go test -cover ./...
ok      github.com/transparency-dev/tesseract   0.026s  coverage: 64.9% of statements
        github.com/transparency-dev/tesseract/cmd/aws           coverage: 0.0% of statements
        github.com/transparency-dev/tesseract/cmd/experimental/migrate/gcp              coverage: 0.0% of statements
        github.com/transparency-dev/tesseract/cmd/gcp           coverage: 0.0% of statements
ok      github.com/transparency-dev/tesseract/internal/client   0.024s  coverage: 59.6% of statements
        github.com/transparency-dev/tesseract/internal/client/gcp               coverage: 0.0% of statements
ok      github.com/transparency-dev/tesseract/internal/ct       (cached)        coverage: 86.2% of statements
ok      github.com/transparency-dev/tesseract/internal/hammer   0.525s  coverage: 27.2% of statements
ok      github.com/transparency-dev/tesseract/internal/hammer/loadtest  0.664s  coverage: 7.1% of statements
ok      github.com/transparency-dev/tesseract/internal/lax509   0.114s  coverage: 81.9% of statements
        github.com/transparency-dev/tesseract/internal/otel             coverage: 0.0% of statements
ok      github.com/transparency-dev/tesseract/internal/testonly/storage/posix   (cached)        coverage: 87.5% of statements
        github.com/transparency-dev/tesseract/internal/types/rfc6962            coverage: 0.0% of statements
        github.com/transparency-dev/tesseract/internal/types/staticct           coverage: 0.0% of statements
ok      github.com/transparency-dev/tesseract/internal/types/tls        0.020s  coverage: 98.8% of statements
ok      github.com/transparency-dev/tesseract/internal/x509util 0.031s  coverage: 53.7% of statements
        github.com/transparency-dev/tesseract/storage           coverage: 0.0% of statements
        github.com/transparency-dev/tesseract/storage/aws               coverage: 0.0% of statements
        github.com/transparency-dev/tesseract/storage/gcp               coverage: 0.0% of statements
```

roger2hk:
Latest coverage data:

```sh
[...]

PR #182, opened 2025-03-13
Add wrong signer type and empty signer test cases into `TestNewLog`

roger2hk:
Towards https://github.com/transparency-dev/static-ct/issues/103.


internal/scti/ctlog_test.go
```diff
@@ -3,21 +3,33 @@ package scti
...
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"os"
 	"strings"
 	"testing"
 
-	"github.com/google/trillian/crypto/keys/pem"
+	trillianpem "github.com/google/trillian/crypto/keys/pem"
```
mhutchinson, line 17: I thought you were trying to remove dependencies on trillian from TesseraCT?

roger2hk: Correct. This pull request needs the "encoding/pem" module. I will remove the "github.com/google/trillian/crypto/keys/pem" dependency in the another pull request. Created https://github.com/transparency-dev/static-ct/issues/183 to track.

mhutchinson (approved):
I'm still unclear on why we now have 2 methods for opening private keys, but if the trillian dep will be removed then I'll wait for the next episode to find out.

PR #186, opened 2025-03-14
Update test keys/certs gen tool doc comment

roger2hk:


internal/testdata/gen/gen.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// gen is a tool to generate test EC keys and test certificates for hammer.
+// gen is a tool to generate test EC keys and test certificates.
```
phbnf, line 15: nit, RSA as well

phbnf (approved):

internal/testdata/gen/gen.go
roger2hk: Good catch!

Issue #104, opened 2025-02-06
[GCP] Launch a public static-ct-api test log


phbnf:
- [x] bring up a staging GCP project
- [x] bring up a log in there
- [x] preload the log from argon2025h1
- [x] map url to log
- [x] make the log public

PR #192, opened 2025-03-14
organize terraform configs

phbnf:
Towards #104  

This PR Reorganize modules to group cloudbuild configs together, and tesseract configs together. This is going to be useful once we have more than one instance each. It's already the case for tesseract since we have the test tesseract config and the conformance once. A followup PR will add more of cloudbuild for the staging environment.

On `static-ct`:
```
$ terragrunt plan
....
No changes. Your infrastructure matches the configuration.
```


deployment/live/gcp/static-ct/logs/ci/README.md
roger2hk, line 1: The path in L50 in the Manual Deployment section needs to be updated.

roger2hk (approved):

Issue #44, opened 2024-11-08
Drop dependency on c-t-go x509 and asn1 forks


AlCutter:
The original CTFE implementation contained forks of Go's `x509` and `asn1` packages, at the time this was necessary in order to make them more permissive in what they'd accept. This has become considerably less important now that tools like cablint have been adopted by the industry.

Maintaining these forks comes with a) significant work to keep them up-to-date, or b) risk that not updating them leaves them vulnerable to known issues, and/or they [otherwise rot](https://github.com/google/certificate-transparency-go/issues/1606).

We should use this opportunity to switch [back] to Go's stdlib `x509` and `asn1` packages.

PR #198, opened 2025-03-18
Migrate away from c-t-go/x509 and c-t-go/asn1

phbnf:
Towards #44 

This PR migrates away from `certificate-transparency-go/x509` and `certificate-transparency-go/asn1`. A follow up PR will drop dependency on certificate-transparency-go/tls.

To allow for such a transition, this PR does a few things:
 - modifies how certificates and their fields are printed: `c-t-go/x509` has a lot of methods to print certificates and their fields. `crypto/x509` does not have methods to print full certs anymore, but individual fields that implement the stringer interface can be printed. We only every used full certificate printing in `chain_validation_test.go`, so I've simply replaced this with printing its md5sum for now. This allows to delete the full certificate printing methods. If we need them again, we can use other libraries.
 - moves `BuildPrecertTBS` from `c-t-go/x509` to `x509util/ct`
 - changes how `isPreIssuer` works, since the `crypto/x509` library won't parse CT extended key usage extensions.


internal/scti/chain_validation.go
```diff
@@ -225,32 +227,18 @@ func validateChain(rawChain [][]byte, validationOpts ChainValidationOpts) ([]*x5
...
-		DisablePathLenChecks:        true,
-		DisableNameConstraintChecks: true,
-		DisableNameChecks:           false,
-		KeyUsages:                   validationOpts.extKeyUsages,
+	// We can now do the verification. Use x509fork with looser verification
+	// constraints to:
+	//  - allow pre-certificates and chains with pre-issuers
+	//  - allow certificate without policing them since this is not CT's responsibility
+	// See /internal/x509fork/README.md for further information.
+	verifyOpts := x509fork.VerifyOptions{
```
AlCutter, line 235: Naming idea for this fork package: `lax509` :)

internal/x509fork/cert_pool.go
```diff
@@ -234,9 +234,7 @@ func (s *CertPool) AppendCertsFromPEM(pemCerts []byte) (ok bool) {
 
 // Subjects returns a list of the DER-encoded subjects of
 // all of the certificates in the pool.
-//
-// Deprecated: if s was returned by [SystemCertPool], Subjects
-// will not include the system roots.
+// Undeprecated in this fork.
```
roger2hk, line 237: It would be weird to have only "Undeprecated in this fork" after merging this PR. Can we add a link to the original package?

https://pkg.go.dev/crypto/x509#CertPool.Subjects

internal/scti/chain_validation_test.go
```diff
@@ -690,7 +692,7 @@ func TestPreIssuedCert(t *testing.T) {
 				t.Fatalf("failed to ValidateChain: %v", err)
 			}
 			for i, c := range chain {
-				t.Logf("chain[%d] = \n%s", i, x509util.CertificateToString(c))
+				t.Logf("chain[%d] = \n%s", i, md5.Sum(c.Raw))
```
AlCutter, line 695: md5?

Would printing the `Subject` make it easier to tie back to the certs in the chain?

internal/scti/handlers.go
```diff
@@ -510,7 +509,13 @@ func entryFromChain(chain []*x509.Certificate, isPrecert bool, timestamp uint64)
...
-	return slices.Contains(issuer.ExtKeyUsage, x509.ExtKeyUsageCertificateTransparency)
+func isPreIssuer(cert *x509.Certificate) bool {
+	// Look for the extension in the Extensions field and not ExtKeyUsage
+	// since crypto/x509 does not recognize this extension as an ExtKeyUsage.
+	for _, ext := range cert.Extensions {
+		if types.OIDExtKeyUsageCertificateTransparency.Equal(ext.Id) {
+			return true
+		}
+	}
+	return false
```
AlCutter, line 520: I've seen this in a couple of places, feels like it might reasonably live exported in `x509util`?

internal/x509util/ct.go
```diff
@@ -0,0 +1,193 @@
...
+	"math/big"
+	"time"
+)
+
+var (
+	oidExtensionAuthorityKeyId = asn1.ObjectIdentifier{2, 5, 29, 35}
+	// These extensions are defined in RFC 6962 s3.1.
+	oidExtensionCTPoison                        = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 3}
+	oidExtensionKeyUsageCertificateTransparency = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 4}
+)
```
AlCutter, line 31: could/should these live next to the other RFC6962 OIDs currently in `types`?

internal/x509util/ct.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	extAt := -1
+	for i, ext := range tbs.Extensions {
+		if ext.Id.Equal(oid) {
+			if extAt != -1 {
+				return nil, errors.New("multiple extensions of specified type present")
+			}
+			extAt = i
+		}
+	}
```
AlCutter, line 77: I think you may be able to use `slices.IndexFunc` here?

internal/x509util/ct.go
```diff
@@ -0,0 +1,193 @@
...
+			return nil, fmt.Errorf("issuer does not have CertificateTransparency extended key usage")
+		}
+
+		keyAt := -1
+		for i, ext := range tbs.Extensions {
+			if ext.Id.Equal(oidExtensionAuthorityKeyId) {
+				keyAt = i
+				break
+			}
+		}
```
AlCutter, line 157: might be able to use `slices.IndexFunc` here?

AlCutter (approved):
Tangential thought; is `types` really two separate packages, e.g. `rfc6962` and `staticct`?

internal/x509util/ct.go
```diff
@@ -0,0 +1,193 @@
...
+		// The x509 package does not parse CT EKU, so look for it in
+		// extensions directly.
+		seenCTEKU := false
+		for _, ext := range preIssuer.Extensions {
+			if ext.Id.Equal(oidExtensionKeyUsageCertificateTransparency) {
+				seenCTEKU = true
+				break
+			}
+		}
+		if !seenCTEKU {
```
roger2hk, line 147: Could `slices.ContainsFunc` be used here?

roger2hk (approved):

internal/scti/handlers.go
phbnf: Yeah - there's a TODO to merge them, I'll do this in a followup PR. This PR was 90% done before the other `isPreissuer` was added.

internal/x509util/ct.go
phbnf: They could, yes: but what's the benefit?

I decided to leave them here because:
 - we're only talking about two vars, defined in the RFC that and haven't changed in 10+ years
 - it makes this file more readable
 - `x509util` is really what it says: it's a util for `x509` and it does not import anything else, it just depends on standard libraries.

internal/x509util/ct.go
phbnf: I can yeah. I'll leave it for now in this PR since these methods are being copied from c-t-go, and can send a followup PR.

I find that the end result is not necessarily easier to read though, especially here since we need to use the function twice to find duplicate extensions. The current implementation only requires one loop iteration, this one two. Plus, the base implementation of `slices.Index` `slices.IndexFunc` aren't really bringing any magic. I've done a fair amount of function programming before, and I really really really like and the `slice` package, but for very simple usecases like this I find it way less readable.

internal/x509util/ct.go
phbnf: Ack - will do in a followup PR, this PR is trying no to change the implementation of these functions too much and to copy them over from c-t-go.

internal/x509util/ct.go
phbnf: Ack - will do in a followup PR, this PR is trying no to change the implementation of these functions too much and to copy them over from c-t-go.

internal/scti/chain_validation.go
phbnf: Sending a different PR to avoid making this one even larger.

PR #200, opened 2025-03-19
remove objects and structs we don't need

phbnf:
Towards #44.

I'll pay a lot of attention to make sure that the integration tests pass after submitting this, and ensure I'm not deleting important marshaling methods.


AlCutter (approved):
<3

PR #213, opened 2025-03-24
Change metric naming, unit, and http fields

phbnf:
Towards #105.

Following recommendations from: https://prometheus.io/docs/practices/naming/

Prefix all the metrics with `ct_`, and suffix with units. Also, use seconds for the SCT timestamp unit since this is what guidelines recommend.... even if an SCT timestamp is in milliseconds... I'm a bit on the fence for this one, but these metrics are meant for graphs and human consumption, so seconds seem to more adapted indeed (and for anything more fancy, milliseconds are still available since the number exported is a float).

Note that I'm explicitly specifying the `ct_` prefix manually and not using a wrapper, not to break code searching: I find that searching for a metric name is very useful when investigating an outage.

### DIFF
--- a/internal/scti/handlers.go
+++ b/internal/scti/handlers.go
@@ -68,7 +68,7 @@ var (
 	lastSCTTimestamp *prometheus.GaugeVec     // origin => value
 	reqsCounter      *prometheus.CounterVec   // origin, op => value
 	rspsCounter      *prometheus.CounterVec   // origin, op, code => value
-	rspDuration      *prometheus.HistogramVec // origin, op, code => value
+	reqsDuration     *prometheus.HistogramVec // origin, op, code => value
 )
 
 // setupMetrics initializes all the exported metrics.
@@ -98,16 +98,16 @@ func setupMetrics() {
 			Help: "Number of requests",
 		},
 		[]string{"origin", "operation"})
+	reqsDuration = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "ct_http_request_duration_seconds",
+			Help: "Latency of responses in seconds",
+		},
+		[]string{"origin", "operation", "code"})
 	rspsCounter = promauto.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "ct_http_responses_total",
 			Help: "Number of responses",
-		},
-		[]string{"origin", "operation", "code"})
-	rspDuration = promauto.NewHistogramVec(
-		prometheus.HistogramOpts{
-			Name: "ct_http_request_duration_seconds",
-			Help: "Latency of responses in seconds",
 		},
 		[]string{"origin", "operation", "code"})
 }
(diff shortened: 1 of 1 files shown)


internal/scti/handlers.go
```diff
@@ -68,48 +68,48 @@ var (
 	lastSCTTimestamp *prometheus.GaugeVec     // origin => value
 	reqsCounter      *prometheus.CounterVec   // origin, op => value
 	rspsCounter      *prometheus.CounterVec   // origin, op, code => value
-	rspLatency       *prometheus.HistogramVec // origin, op, code => value
+	rspDuration      *prometheus.HistogramVec // origin, op, code => value
```
AlCutter, line 71: Out of curiosity, what's the thinking behind the name change? 
What I think this metric represents is called `latency distribution` in, er, other automatic monitoring systems we're intimately familiar with :)

If you're intent on changing it, "response duration" sounds a bit odd to my ears, but how about `rspTime`? Also, the description of the metric still explains it in terms of `Latency of responses`, so I guess that needs updating to match.

internal/scti/handlers.go
```diff
@@ -68,48 +68,48 @@ var (
...
-			Name: "known_logs",
+			Name: "ct_known_logs_info",
 			Help: "Set to 1 for known logs",
 		},
 		[]string{"origin"})
 	lastSCTTimestamp = promauto.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "last_sct_timestamp",
-			Help: "Time of last SCT in ms since epoch",
+			Name: "ct_sct_last_timestamp_seconds",
```
AlCutter, line 85: RFC6962 specifies `timestamp` in milliseconds, so it feels a bit unusual to export a metric in fractional seconds.
I'm sure there's a good reason, so it would be worth a comment somewhere in here explaining why we're doing that rather than exporting the value in its natural form.

internal/scti/handlers.go
```diff
@@ -68,48 +68,48 @@ var (
...
+			Name: "ct_http_responses_total",
 			Help: "Number of responses",
 		},
-		[]string{"origin", "op", "code"})
-	rspLatency = promauto.NewHistogramVec(
+		[]string{"origin", "operation", "code"})
+	rspDuration = promauto.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name: "http_latency",
+			Name: "ct_http_request_duration_seconds",
```
AlCutter, line 109: nit: variable naming is in terms of "response", but metric name is now in terms of "request" - would be good to keep them consistent if possible.

internal/scti/handlers.go
phbnf: Apologies, this PR lacked something very important: a link to https://prometheus.io/docs/practices/naming/, I've now added it. I'm 100% onboard with your comments and this looked a bit odd to me at first sight, being used to other monitoring systems... but it seems to be the prometheus way.

internal/scti/handlers.go
phbnf: Same here, I forgot to link to https://prometheus.io/docs/practices/naming/, I've now added it. As explained in the PR overview comment, I'm on the fence as well, but I'm following best practices. I can easily be convinced that we should not follow that recommendation because of the RFC :D. I thought I'd give it a try though since in practice, I could not justify to myself that having milliseconds would actually be useful, other than matching with the RFC. This metric is likely to be used for two things:
 - graphs: either we just want to see a graph going upright, or we'll change the axis to be human parsable with a human readable time representation
 - alerting: if any it's unlikely we create alerts at the millisecond level

internal/scti/handlers.go
phbnf: Oh yeah, good point!

internal/scti/handlers.go
AlCutter: I think the intent of the naming doc you linked was the `http_` prefix there, and the `..._request_duration_seconds` was just an example.

I found this "counter" (_hah!_) example in the [histogram section](https://prometheus.io/docs/tutorials/understanding_metric_types/#histogram) of the prom docs: 
> A common example would be the time it takes to reply to a request, called latency.

So I think there's latitude to fit in with styles we're familiar with.

Issue #110, opened 2025-02-06
[GCP] static-ct-api log ready for alpha


phbnf:
This is a parent bug.

Issue #109, opened 2025-02-06
[AWS] static-ct-api log ready for alpha


phbnf:
This is a parent bug.

PR #216, opened 2025-03-26
Add Secret(s) Manager security risk warning messages

roger2hk:
Towards #109, #110.

### DIFF
--- a/deployment/modules/aws/tesseract/test/main.tf
+++ b/deployment/modules/aws/tesseract/test/main.tf
@@ -11,6 +11,9 @@ module "storage" {
   ephemeral   = var.ephemeral
 }
 
+# [WARNING]
+# Using secrets manager does NOT guarantee the security of the signing keys. 
+# There are significant security risks if the secrets are mismanaged.
 module "secretsmanager" {
   source = "../../secretsmanager"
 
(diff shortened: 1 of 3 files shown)


phbnf:
This message isn't explicit about what's actually happening. I think it's important to be explicit to prevent someone from copy-pasting this config to bring up their production log. Here's an alternative suggestion:

```
# [WARNING]
# This module will hardcode unencrypted private keys in the Terraform state file.
# DO NOT use this for production logs.
```

Let's also put this "WARNING" banner in:
  - the secretmanager modules `main.tf` files
  - a README.md file in the secretmanager directories?

I really really want to make sure nobody uses this for a production log, hence the request for extra warning messages. I've created #219 to track implementation of a solution for production logs.

phbnf:
Let's add all the messages for now since that's the state of the world, and remove them once we've designed a better process to bring up new logs safely?
I'm not sure we'll get to do this before the Alpha release. If we do, that's great, and we can remove them. If we don't, I'd rather err on the safe side for now.

roger2hk:
Ah, I've already got that done for AWS in https://github.com/transparency-dev/static-ct/pull/220 when I saw this message.

roger2hk:
Closing this as the warning messages will be covered in #220 and #222.

Issue #219, opened 2025-03-26
Provision key in a safe way


phbnf:
Currently, both the AWS and GCP secret manager implementation store unencrypted private keys in the Terraform state file. This is ok for test logs, but we need a different solution for production logs.

roger2hk:
Note that AWS/GCP Secret(s) Manager has already encrypted the secrets with the corrsponding KMS key(s). The insecure part is the `hashicorp/tls` resource.

Extracting the `tls_private_key` into its own module and naming the module as "unsafe/insecure" could greatly avoid the mis-use of `hashicorp/tls` resource.

phbnf:
Yes, 100%. I'm thinking as well that we could carve out the module that creates the private key, and provide instructions about how to deploy this module safely and smoothly to have a secure new log creation customer journey.

phbnf:
I'll mark this as fixed. On GCP, we can now use the new tools that @AlCutter added to generate keys and store them in Secret Manager. The GCP binary can read them. 

On POSIX the `private_key` file seem to have been working well for IPng and TrustAsia folks.

On AWS, the binary can read keys from Secrets Manager. Someone using an AWS log in production is better suited than us to write a tool to generate and store keys appropriately.

PR #222, opened 2025-03-26
[GCP] Extract `tls_private_key` resource from `hashicorp/tls` provider into an insecure module

roger2hk:
Towards https://github.com/transparency-dev/static-ct/issues/110, https://github.com/transparency-dev/static-ct/issues/219.


deployment/modules/gcp/insecuretlskey/README.md
```diff
@@ -0,0 +1,3 @@
+# [WARNING]
+# This module will hardcode unencrypted private keys in the Terraform state file.
```
AlCutter, line 2: /hardcode/store/?

deployment/modules/gcp/tesseract/test/main.tf
```diff
@@ -14,5 +14,14 @@ module "storage" {
...
   source = "../../secretmanager"
 
-  base_name = var.base_name
+  base_name                                  = var.base_name
+  tls_private_key_ecdsa_p256_public_key_pem  = module.insecuretlskey.tls_private_key_ecdsa_p256_public_key_pem
+  tls_private_key_ecdsa_p256_private_key_pem = module.insecuretlskey.tls_private_key_ecdsa_p256_private_key_pem
+}
+
+# [WARNING]
+# This module will hardcode unencrypted private keys in the Terraform state file.
```
AlCutter, line 23: /hardcode/store/?

AlCutter (approved):

deployment/modules/gcp/insecuretlskey/README.md
roger2hk: Done.

phbnf:
I like that the bit that we know for sure is insecure is carved out. However, the private key is still passed as an argument, and I don't know for sure how safe this is, and I wouldn't be surprised if there were ways to extract the key from debug logs, or by printing it. Until we have a battle and tested production I think we'd still want to have warnings that mention that the secretmanager module is not safe? What do you think?

roger2hk:
> I like that the bit that we know for sure is insecure is carved out. However, the private key is still passed as an argument, and I don't know for sure how safe this is, and I wouldn't be surprised if there were ways to extract the key from debug logs, or by printing it. Until we have a battle and tested production I think we'd still want to have warnings that mention that the secretmanager module is not safe? What do you think?

Gotcha, I understand your concern. Enforcing the write-only attributes (i.e. `secret_data` to `secret_data_wo`) for the private key field can mitigate the risk. All arguments marked as write-only values will not be stored in the Terraform state. There will also be no way to retrieve the private key field secret data value.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/using_write_only_attributes

Would that address your concern?

phbnf:
I did not know about this! It looks promising indeed, and it seems to be supported by AWS as well. I had an alternative idea in mind, which would not depend on Terraform. It's not a full design, and would certainly requite a bit more work.

Throughout the lifecycle of a log, we need to perform these operations:
 1. Generate the private and public key: today, this happens with terraform
 2. Store the private key (or a secret that allows to generate/access it) in Secret(s) Manager: today, this happens via terraform as well
 3. Load the private key in TesseraCT's memory: this happens using GCP Go APIs directly

We could write a simple go binary that achieves 1, 2 using GCP Go APIs. A user would be set up in advance, with the right permissions to launch this binary, and run it once only. That would make 1. and 2. more transparent I think (a few lines of Go), and means that all key access happens through Go APIs (like 3.). I like that the key provisioning system would have a low number of dependencies, that it would be easier to read the code that generates it, and easy to track where the key lives (only in memory and Secrets Manager).

This means that Terraform would never touch keys, and that the users allowed to deploy Terraform wouldn't have access to it either. Depending on the access that the user has, we could even make sure that key generation can only happen on compute ressources that run on GCP, not on someone's laptop for instance. The only thing we'd need in terraform would be a way to identify the key, through a name or an id for instance, which could be hardcoded for every log. It comes with the downside that one piece of infrastructure (keys) would be created without Terraform, but I think that's an ok compromise.

Note that the binary that generates this key could still be launch via terraform to make the deployment process smooth!
I'm not 100% sure yet how role delegation would work though, I'd need to look into it a bit more to see how do that safely for production logs, but I'd be surprised if there was no way to do this safely. For test logs, it shouldn't be an issue.

phbnf:
In the meantime, if the write_only feature makes the setup a bit better, sure thing! Would it be compatible with this PR, or would we have to go back to using a single module, specifically because of the write_only bit? We could scope this to the private key only, right?

phbnf:
Does this forbid the use of opentofu?

roger2hk:
> Does this forbid the use of opentofu?

It should be ready now. Please try again to see if it works in OpenTufu.

roger2hk:
New warning message after bumping the GCP provider to the latest version. We can also move the public key to the write-only attribute, and then grab the secret value from `gcloud secrets versions access VERSION_ID --secret=SECRET_ID`.

```
17:14:28.879 STDOUT terraform: │ Warning: Available Write-only Attribute Alternative
17:14:28.879 STDOUT terraform: │ 
17:14:28.880 STDOUT terraform: │   with module.secretmanager.google_secret_manager_secret_version.sctfe_ecdsa_p256_public_key,
17:14:28.880 STDOUT terraform: │   on ../../secretmanager/main.tf line 34, in resource "google_secret_manager_secret_version" "sctfe_ecdsa_p256_public_key":
17:14:28.880 STDOUT terraform: │   34:   secret_data = var.tls_private_key_ecdsa_p256_public_key_pem
17:14:28.880 STDOUT terraform: │ 
17:14:28.880 STDOUT terraform: │ The attribute secret_data has a write-only alternative secret_data_wo
17:14:28.880 STDOUT terraform: │ available. Use the write-only alternative of the attribute when possible.
17:14:28.880 STDOUT terraform: │ 
17:14:28.880 STDOUT terraform: │ (and one more similar warning elsewhere)
```

phbnf:
Last we looked into this with @roger2hk , this turned out to be more complex than we initially thought because OpenTofu and Terraform did not provide the same features. It's hard to know what is really safe to generate keys using Terraform, and we'll do things in a different way in #219.

PR #227, opened 2025-03-31
Bump golangci-lint to 2.0.2

roger2hk:
Unblock https://github.com/transparency-dev/static-ct/pull/225.

### DIFF
--- a/internal/scti/handlers.go
+++ b/internal/scti/handlers.go
@@ -305,7 +305,7 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
 		if err != nil {
 			if errors.Is(err, tessera.ErrPushback) {
 				w.Header().Add("Retry-After", "1")
-				return http.StatusServiceUnavailable, fmt.Errorf("Tessera sequencer pushed back: %v", err)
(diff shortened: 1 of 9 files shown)


internal/scti/handlers.go
```diff
@@ -305,7 +305,7 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
 		if err != nil {
 			if errors.Is(err, tessera.ErrPushback) {
 				w.Header().Add("Retry-After", "1")
-				return http.StatusServiceUnavailable, fmt.Errorf("Tessera sequencer pushed back: %v", err)
+				return http.StatusServiceUnavailable, fmt.Errorf("pushed back from Tessera sequencer: %v", err)
```
phbnf, line 308: nit, received pushback from Tessera sequencer. Also, we should potentially revisit this: this reveals information about the internals of the server, but doesn't go through `sendHTTPError`.

phbnf (approved):

internal/scti/handlers.go
roger2hk: Updated to "received pushback from Tessera sequencer".

Right, the `MaskInternalErrors` option is being ignored here.

roger2hk: Created https://github.com/transparency-dev/static-ct/issues/229 to follow up later.

PR #228, opened 2025-03-31
Add AWS performance doc

roger2hk:
Towards https://github.com/transparency-dev/static-ct/issues/108.

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -2,7 +2,7 @@
 
 TesseraCT is designed to meet current CT issuance load in a cost-effective manner.
 
-The indicative figures below were measured using the [CT hammer tool](/internal/hammer/) as of [commit `2872ea2`](https://github.com/transparency-dev/static-ct/commit/2872ea2387b2d3077eb832112277eb19a7a907bd). The performance tests were conducted in a controlled environment with deterministic synthetic certificates for a limited amount of time. QPS was measured using the average values collected over the test period.
+The performance tests were conducted in a controlled environment with deterministic synthetic certificates for a limited amount of time. QPS was measured using the average values collected over the test period.
 
 > [!NOTE]
 > These are not definitive numbers, and that more tests are to come with an improved codebase.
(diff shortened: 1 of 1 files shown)


docs/performance.md
```diff
@@ -129,4 +131,50 @@ MiB Swap:      0.0 total,      0.0 free,      0.0 used.   2989.2 avail Mem
...
 
-Coming soon.
+The indicative figures below were measured using the [CT hammer tool](/internal/hammer/) as of [commit `2940bba`](https://github.com/transparency-dev/static-ct/commit/2940bba60a49bad1b78ae8f2ded2c893b1b133ad). 
+
+#### t3a.small EC2 Instance + Aurora MySQL db.r5.large
+
+- t3a.small (2 vCPUs, 2 GB Memory)
+  - General Purpose SSD (gp3)
+    - IOPS: 3000
+    - Throughput: 125
```
AlCutter, line 141: Is there a unit to go with this `Throughput` number?

AlCutter (approved):

docs/performance.md
roger2hk: Added, thanks!

phbnf (approved):

phbnf:
I'm thinking maybe we could do additional tests with: better disk IO and a smaller database? What do you think?

roger2hk:
> I'm thinking maybe we could do additional tests with: better disk IO and a smaller database? What do you think?

Sure. I will come back to this after building the CI pipeline.

AlCutter:
@roger2hk does 3KiB I/O make sense as an IOPS measurement?
I think IOPS are the unit for that stat.

roger2hk:
> @roger2hk does 3KiB I/O make sense as an IOPS measurement? I think IOPS are the unit for that stat.

IOPS is indeed the unit. `(3KiB I/O)` represents the required I/O size to reach maximum IOPS within the volume's throughput limit. I can add a footnote or remove that if you find it misleading/confusing.

https://docs.aws.amazon.com/ebs/latest/userguide/ebs-volume-types.html#vol-type-ssd

AlCutter:
> > @roger2hk does 3KiB I/O make sense as an IOPS measurement? I think IOPS are the unit for that stat.
> 
> IOPS is indeed the unit. `(3KiB I/O)` represents the required I/O size to reach maximum IOPS within the volume's throughput limit. I can add a footnote or remove that if you find it misleading/confusing.
> 
> https://docs.aws.amazon.com/ebs/latest/userguide/ebs-volume-types.html#vol-type-ssd

Ah, yes a link or adding "With" or "Using" just before the "3KiB I/O" would help - thanks!

PR #234, opened 2025-04-04
split types into rfc6962 and staticctapi

phbnf:
Addressing https://github.com/transparency-dev/static-ct/pull/198#pullrequestreview-2694135688


internal/hammer/loadtest/workers.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/transparency-dev/merkle/proof"
 	"github.com/transparency-dev/merkle/rfc6962"
 	"github.com/transparency-dev/static-ct/internal/client"
-	"github.com/transparency-dev/static-ct/internal/types"
+	rfc69621 "github.com/transparency-dev/static-ct/internal/types/rfc6962"
```
roger2hk, line 32:
```suggestion
	"github.com/transparency-dev/static-ct/internal/types/rfc6962"
```

internal/hammer/loadtest/workers.go
```diff
@@ -246,7 +246,7 @@ func (w *LogWriter) Run(ctx context.Context) {
 
 		// TODO: Remove the json.Unmarshal by generating the chain and
 		// marshaling the add chain request from w.gen() at a later stage.
-		var req types.AddChainRequest
+		var req rfc69621.AddChainRequest
```
roger2hk, line 249:
```suggestion
		var req rfc6962.AddChainRequest
```

internal/hammer/loadtest/workers.go
```diff
@@ -449,7 +449,7 @@ func isPreIssuer(cert *x509.Certificate) bool {
 	// Look for the extension in the Extensions field and not ExtKeyUsage
 	// since crypto/x509 does not recognize this extension as an ExtKeyUsage.
 	for _, ext := range cert.Extensions {
-		if types.OIDExtKeyUsageCertificateTransparency.Equal(ext.Id) {
+		if rfc69621.OIDExtKeyUsageCertificateTransparency.Equal(ext.Id) {
```
roger2hk, line 452:
```suggestion
		if rfc6962.OIDExtKeyUsageCertificateTransparency.Equal(ext.Id) {
```

roger2hk (approved):

internal/types/staticctapi/staticctapi.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package types
+package staticctapi
```
AlCutter, line 15: `staticctapi` is a bit "letter soupy", how about just `staticct` (or `static-ct`)?
Should be ok particularly since we're planning to rename this repo to `tesseract`...

AlCutter (approved):

internal/hammer/loadtest/workers.go
phbnf: Good catch. I can't do this because of a different import, but I sorted this out.

PR #242, opened 2025-04-08
Bump Tessera & switch to OpenTelemetry

AlCutter:
This PR bumps the `tessera` dep to HEAD, and updates monitoring in this package to adopt [OpenTelemetry](https://opentelemetry.io) as used in Tessera.

Existing prometheus metric have been migrated to OTel, and a few initial top-level trace spans added to show how it works. More metrics & traces could be added as the code stablises.

The GCP binary has been updated to auto-export telemetry to Cloud Monitoring.

For now, I've removed the prometheus metrics exporting from the AWS binary and OTel export support will need to be added at a later date.


internal/scti/handlers.go
```diff
@@ -63,53 +62,38 @@ var (
...
-			Name: "http_latency",
-			Help: "Latency of responses in seconds",
-		},
-		[]string{"origin", "op", "code"})
+	knownLogs = mustCreate(meter.Int64Gauge("tesseract.known_logs",
+		metric.WithDescription("Set to 1 for known logs")))
+
+	lastSCTTimestamp = mustCreate(meter.Int64Gauge("tesseract.last_sct.timestamp",
+		metric.WithDescription("Time of last SCT since epoch"),
+		metric.WithUnit("ms")))
```
phbnf, line 81: (To avoid ambiguity with my other comment, let's keep ms here since this is the unit in the RFC)

internal/scti/handlers.go
```diff
@@ -63,53 +62,38 @@ var (
...
+
+	lastSCTIndex = mustCreate(meter.Int64Gauge("tesseract.last_sct.index",
+		metric.WithDescription("Index of last SCT"),
+		metric.WithUnit("{entry}")))
+
+	reqsCounter = mustCreate(meter.Int64Counter("tesseract.http_request.count",
```
phbnf, line 87: I think the unit should be "request" here? https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserveractive_requests

internal/scti/handlers.go
```diff
@@ -63,53 +62,38 @@ var (
...
+		metric.WithDescription("CT HTTP requests")))
+
+	rspsCounter = mustCreate(meter.Int64Counter("tesseract.http_response.count",
+		metric.WithDescription("CT HTTP responses")))
+
+	rspLatency = mustCreate(meter.Float64Histogram("tesseract.http_response.duration",
+		metric.WithDescription("CT HTTP response duration"),
+		metric.WithExplicitBucketBoundaries(otel.LatencyHistogramBuckets...),
+		metric.WithUnit("ms")))
```
phbnf, line 96: Let's use seconds here as a unit: it's Opentelemetry's default unit for time. I think we should also rename this metric to `http_request.duration` rather, to align with https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration. I can't tell for sure, but should this be `tesseract.http.request.count` rather?

In all fairness: If I were to write this from scratch, I'd use latency... and if I could not use latency I would use "response.duration". But it looks like people who've thought about this for some time have made other decisions, let's stick to default and embrace Opentelemtry way of doing things.

internal/scti/handlers.go
```diff
@@ -63,53 +62,38 @@ var (
...
-			Name: "last_sct_index",
-			Help: "Index of last SCT",
-		},
-		[]string{"origin"})
-	reqsCounter = promauto.NewCounterVec(
-		prometheus.CounterOpts{
-			Name: "http_reqs",
-			Help: "Number of requests",
-		},
-		[]string{"origin", "ep"})
```
phbnf, line 100: Do I gather correctly that we can't specify attributes at metric creation time?

internal/scti/otel.go
```diff
@@ -0,0 +1,41 @@
...
+
+const name = "github.com/transparency-dev/static-ct/internal/scti"
+
+var (
+	meter  = otel.Meter(name)
+	tracer = otel.Tracer(name)
+)
+
+var (
+	codeKey      = attribute.Key("tesseract.code")
```
phbnf, line 31: I think this should be `status_code` from https://opentelemetry.io/docs/specs/semconv/http/http-metrics/

phbnf (approved):

phbnf (approved):

PR #246, opened 2025-04-08
Print Terraform output in separate step

phbnf:
This will ensure that if `terraform apply` fails, the cloudbuild step will also fail.

We need to do the same thing for arch2025h1, it will come in a few PR, it's somewhere in my stack, but would be tricky to rebase since I'm also renaming a few files.

### DIFF
--- a/deployment/modules/gcp/cloudbuild/conformance/main.tf
+++ b/deployment/modules/gcp/cloudbuild/conformance/main.tf
@@ -135,7 +135,7 @@ resource "google_cloudbuild_trigger" "build_trigger" {
       wait_for = ["preclean_env", "docker_push_conformance_gcp"]
     }
 
-    ## Print Terragrunt output.
+    ## Print Terragrunt output to files.
     step {
       id     = "terraform_print_output"
       name   = "alpine/terragrunt"
--- a/deployment/modules/gcp/cloudbuild/conformance/main.tf
+++ b/deployment/modules/gcp/cloudbuild/conformance/main.tf
@@ -124,6 +124,22 @@ resource "google_cloudbuild_trigger" "build_trigger" {
       name   = "alpine/terragrunt"
       script = <<EOT
         terragrunt --terragrunt-non-interactive --terragrunt-no-color apply -auto-approve -no-color 2>&1
+      EOT
+      dir    = "deployment/live/gcp/static-ct/logs/ci"
+      env = [
+        "GOOGLE_PROJECT=${var.project_id}",
+        "TF_IN_AUTOMATION=1",
+        "TF_INPUT=false",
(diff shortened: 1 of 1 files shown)


deployment/modules/gcp/cloudbuild/conformance/main.tf
```diff
@@ -124,6 +124,22 @@ resource "google_cloudbuild_trigger" "build_trigger" {
...
+    }
+
+    ## Print Terragrunt output.
+    step {
+      id     = "terraform_print_output"
+      name   = "alpine/terragrunt"
+      script = <<EOT
         terragrunt --terragrunt-no-color output --raw conformance_url -no-color > /workspace/conformance_url
         terragrunt --terragrunt-no-color output --raw conformance_bucket_name -no-color > /workspace/conformance_bucket_name
         terragrunt --terragrunt-no-color output --raw ecdsa_p256_public_key_data -no-color > /workspace/conformance_log_public_key.pem
```
roger2hk, line 145: These lines print the output to the files. I would suggest putting these back into the previous step, and then cat those files in this step.

phbnf: If `terragrunt apply` fails, it will return with a status code of `1` (as it should). If this is the last command ran in the cloud build step, then this return code is also the returned status code of the Cloud Build step. That means that if `terraform apply` fails, then the whole cloudbuild step will fail, and we'd see this in the UI.

If `terragrunt apply` is followed by another command, which itself succeed, then the status code is not propagated, and the cloudbuild step looks successful, even if `terragrunt apply` fails. This PR removes these additional commands, such that the output of the cloudbuild step represents the output of `terragrunt apply`

deployment/modules/gcp/cloudbuild/conformance/main.tf
```diff
@@ -124,6 +124,22 @@ resource "google_cloudbuild_trigger" "build_trigger" {
...
+      env = [
+        "GOOGLE_PROJECT=${var.project_id}",
+        "TF_IN_AUTOMATION=1",
+        "TF_INPUT=false",
+        "TF_VAR_project_id=${var.project_id}"
+      ]
+      wait_for = ["preclean_env", "docker_push_conformance_gcp"]
+    }
+
+    ## Print Terragrunt output.
```
roger2hk, line 138:
```suggestion
    ## Print Terragrunt output to files.
```

deployment/modules/gcp/cloudbuild/conformance/main.tf
roger2hk: Oh gotcha. I thought "Print the output" means printing to the console. It totally makes sense to separate them into two steps.

roger2hk (approved):

PR #249, opened 2025-04-09
Remove `metrics_endpoint` flag in GCP CI pipeline

roger2hk:
The GCP CI pipeline is now broken since the `metrics_endpoint` flag was already removed in https://github.com/transparency-dev/static-ct/pull/242.


phbnf:
Oops sorry, I had not seen this PR. I ran into the same issue, and did not realized you had raised it already.

PR #250, opened 2025-04-09
Rename terraform directories

phbnf:
Towards #104 

This PR flattens out the live cloudbuild directory in `static-ct-staging`, as suggested by @roger2hk in a previous PR.
It also renames the `preloaded` module to `tesseract`: the preloader won't be called form this module in the end.


deployment/live/gcp/static-ct-staging/cloudbuild/arche2025h1/terragrunt.hcl
```diff
@@ -0,0 +1,21 @@
...
+
+include "root" {
+  path   = find_in_parent_folders()
+  expose = true
+}
+
+inputs = merge(
+  local,
+  include.root.locals,
+)
```
roger2hk, line 21: Do we need to update this part? It looks like the root terragrunt file is removed.

roger2hk (approved):

Issue #244, opened 2025-04-08
Reorganize methods and functions


phbnf:
Now that we've made good progress on #5, we can move around more things and regroup methods / functions.

phbnf:
We've done quite a lot of refactoring already, we could do more refactoring inside handlers.go, but it's good enough for now.

PR #259, opened 2025-04-11
Make ChaindValidator an interface and use consistent names

phbnf:
Towards #244

### DIFF
--- a/internal/scti/ctlog.go
+++ b/internal/scti/ctlog.go
@@ -24,9 +24,8 @@ type log struct {
 	origin string
 	// signSCT Signs SCTs.
 	signSCT signSCT
-	// chainValidationOpts contains various parameters for certificate chain
-	// validation.
-	chainValidationOpts ChainValidationOpts
+	// chainValidator validates incoming chains.
+	chainValidator ChainValidator
 	// storage stores certificate data.
 	storage Storage
 }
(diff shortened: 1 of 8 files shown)


internal/scti/ctlog.go
```diff
@@ -43,10 +42,18 @@ type Storage interface {
 	GetCertDedupInfo(context.Context, *x509.Certificate) (dedup.SCTDedupInfo, bool, error)
 }
 
+// ChainValidator provides functions to validate incoming chains.
+type ChainValidator interface {
+	Validate(req rfc6962.AddChainRequest, expectingPrecert bool) ([]*x509.Certificate, error)
+	Roots() []*x509.Certificate
+}
```
roger2hk, line 49: There is only one chain validator implementation. Do we really need this interface?

phbnf: Yeah - I was expecting this comment, I thought the same when I opened this PR a days after I wrote it now. It's mostly stylistic, I got quite excited when I saw that a log struct is just a collection of functions {signSCT, storage funcs, root validation funcs}. I could just make the ChainValidator object public, and use this one directly. I don't feel strongly about it.

roger2hk: Okay. Let's get this in.

roger2hk (approved):

roger2hk:
The pull request title mentioned "ChainVerifier" but the code change is "ChainValidator". Shall we update the title?

PR #278, opened 2025-04-17
Use ms buckets

phbnf:
The unit is set to seconds, align buckets with this.


AlCutter (approved):
Maybe the comment (and maybe var name?) could do with updating to be clear these are buckets in second units, but which are usefully granular at the ms to 1-2s scale?

PR #314, opened 2025-05-16
Support `gs://` URI scheme in hammer

roger2hk:
Towards https://github.com/transparency-dev/tesseract/issues/313.

### DIFF
--- a/internal/client/fetcher.go
+++ b/internal/client/fetcher.go
@@ -24,7 +24,6 @@ import (
 	"path"
 	"strings"
 
-	gcs "cloud.google.com/go/storage"
 	"github.com/transparency-dev/tessera/api/layout"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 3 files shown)


AlCutter:
I think adding weird vendor specific URL scheme support to the core of the _client_ is a bit of a slippery slope. How about if we at least put the `GSFetcher` into its own package somewhere, so folks using the client won't depend on that package unless they actively import & use it to create fetchers?

We could also tidy up the hammer API a bit to make it easier for users of that package to decide which fetchers to use - e.g. `loadtest.NewLogClient` could be deleted and replaced with e.g. `loadtest.NewRoundRobinFetcher(f []Fetcher) Fetcher` and `loadtest.NewRoundRobinLeafWriter(w []LeafWriter) LeafWriter`.  This would perhaps fit a bit more naturally with the existing signature of `loadtest.NewHammer` which tries to be agnostic about how fetchers and writers are actually implemented.

wdyt @roger2hk ?

AlCutter:
(e.g. https://github.com/transparency-dev/tessera/pull/647)

roger2hk:
> I think adding weird vendor specific URL scheme support to the core of the _client_ is a bit of a slippery slope. How about if we at least put the `GSFetcher` into its own package somewhere, so folks using the client won't depend on that package unless they actively import & use it to create fetchers?
> 
> We could also tidy up the hammer API a bit to make it easier for users of that package to decide which fetchers to use - e.g. `loadtest.NewLogClient` could be deleted and replaced with e.g. `loadtest.NewRoundRobinFetcher(f []Fetcher) Fetcher` and `loadtest.NewRoundRobinLeafWriter(w []LeafWriter) LeafWriter`. This would perhaps fit a bit more naturally with the existing signature of `loadtest.NewHammer` which tries to be agnostic about how fetchers and writers are actually implemented.
> 
> wdyt @roger2hk ?

Sounds good to me. I've created a `gcp` package and moved `GSFetcher` into it. Thanks for the refactoring in Tessera hammer.

AlCutter (approved):

PR #318, opened 2025-05-19
Add deduplication attribute to metrics

phbnf:
Towards #292
This PR allows each handler to return additional attributes to be set on the monitoring metrics, and uses this to set the duplicate attribute. It's not the cleanest way of doing this, and I'm hoping to refactor the whole handler setup at some point, but it's okay for now.

The same applies for how handlers detect whether or not an entry has been deduped with timestamp, I think we can do better and cleaner, but let's get back to this later.

### DIFF
--- a/internal/ct/handlers.go
+++ b/internal/ct/handlers.go
@@ -287,11 +287,8 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
 		}
 		return http.StatusInternalServerError, nil, fmt.Errorf("couldn't store the leaf: %v", err)
 	}
-	dedupedAttribute := duplicateKey.Bool(false)
 	isDup := dedupedTimeMillis != timeMillis
-	if isDup {
-		dedupedAttribute = duplicateKey.Bool(true)
-	}
+	dedupedAttribute := duplicateKey.Bool(isDup)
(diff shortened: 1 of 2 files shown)


internal/ct/handlers.go
```diff
@@ -271,74 +271,79 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
...
+			return http.StatusServiceUnavailable, nil, fmt.Errorf("received pushback from Tessera sequencer: %v", err)
 		}
-		return http.StatusInternalServerError, fmt.Errorf("couldn't store the leaf: %v", err)
+		return http.StatusInternalServerError, nil, fmt.Errorf("couldn't store the leaf: %v", err)
+	}
+	dedupedAttribute := duplicateKey.Bool(false)
+	isDup := dedupedTimeMillis != timeMillis
+	if isDup {
+		dedupedAttribute = duplicateKey.Bool(true)
 	}
```
roger2hk, line 294:
```suggestion
	isDup := dedupedTimeMillis != timeMillis
	dedupedAttribute := duplicateKey.Bool(isDup)
```

The suggestion improves the code readability for me. See if it's the same for you.

roger2hk (approved):

Issue #321, opened 2025-05-21
Documentation


phbnf:
This bug covers documentation improvements for the Alpha launch.

- [x] landing page
  - [x] overview of repo content (code, main binaries, deployment configs, codelab, auxiliary tools)
- [x] memory utilization considerations
- [x] flags, especially around batch sizes and checkpoint publishing time.
- [x] deployments
   - [x] test deployment to get started fast
   - [ ] good documentation for Google's GCP logs
   - [ ] how to get started to have your own production log
   - [x] Running locally on a VM, and configuring Opentelemtry (otel_project_id + export OTEL_RESOURCE_ATTRIBUTES=="service.name=XX, service.instance.id=X".)
- [x] lax509 and co
- [x] architecture
  - [x] GCP
  - [x] AWS
  - [x] Non-cloud
- [x] how to handle multiple logs
- [x] configuration
- [x] history and evolution from c-t-go/CTFE
- [x] roadmap
- [x] backend compatibility
- [x] List certs that we know won't be parsed with the new lax509 fork

PR #350, opened 2025-06-04
Add batch max size and age flags to AWS and GCP binaries

roger2hk:


roger2hk:
> Would be great to add documentation about how to use this. I added something to #321.

Sure. Documentation is definitely needed. It's very important to understand the performance impact of each flag here. I will add the doc in the upcoming PRs.

PR #359, opened 2025-06-10
Add configuration flags doc

roger2hk:
Towards #321.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,32 @@
+# TesseraCT Configuration
+
+## Flags
+
+### Checkpoint Interval
+
+The `checkpoint_interval` flag controls the interval duration between checkpoint publishing. Tessera enforces the minimum permitted checkpoint interval check during the TesseraCT application initialization process.
+
+| Backend | Minimum Permitted Checkpoint Interval (ms) |
+| ------- | ------------------------------------------ |
+| AWS     | 1000                                       |
+| GCP     | 1200                                       |
+
+### Publication Awaiter
+
+The `enable_publication_awaiter` flag enables the publication awaiter, which waits for a checkpoint larger than the index in the SCT to be published before returning that SCT.
+
+### In-memory Antispam Cache Size
+
+The `inmemory_antispam_cache_size` flags controls the maximum number of entries in the [in-memory antispam cache](https://github.com/transparency-dev/tessera?tab=readme-ov-file#antispam). The value should be calculated against the allocated instance memory size.
+
+### Sequencing Batch
+
+The `batch_max_age` and `batch_max_size` flags control the maximum age and size of entries in a single sequencing batch. Many factors affecting the optimal values for these flags, such as the number of TesseraCT servers, and their steady QPS rate.
+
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -0,0 +1,23 @@
+# TesseraCT Configuration
+
+## Flags
+
+### Publication Awaiter
+
+The `enable_publication_awaiter` flag enables the publication awaiter, which waits for the checkpoint to be published before returning. As the default checkpoint interval is 10 seconds, it is recommended to update the `checkpoint_interval` flag to a lower value for performance. Note that the [maximum rate of writes to the checkpoint file is one write per second on Google Cloud Storage](https://cloud.google.com/storage/quotas#objects). Tessera enforces the minimum permitted checkpoint interval check during the TesseraCT application initialization process.
```
phbnf, line 7: Let's set a default value that works well for CT directly.

phbnf, line 7: Do you want to imply that this flag should be set to value higher than 1 write per second? If yes, say it directly (and we could also implement this in the code).

roger2hk: Sounds good. I would recommend taking the minimum possible values (i.e. 1000ms for AWS and 1200ms for GCP) as the default values. What do you think?

roger2hk: The Tessera code will exit if the flag value is set to lower than 1000ms for AWS and 1200ms for GCP. I can add that here but it could change in the future.

docs/configuration.md
```diff
@@ -0,0 +1,23 @@
...
+
+The `batch_max_age` and `batch_max_size` flags control the maximum age and size of entries in a single sequencing batch. There are many factors affecting the optimal values for these flags. If the `enable_publication_awaiter` flag is `true`, the values of the sequencing batch flags affect the response time.
+
+### AWS
+
+TesseraCT expects both databases from `db_name` and `antispam_db_name` flags are located in the same Aurora DB cluster.
+
+### GCP
+
+The `otel_project_id` flag is only required when running locally on VMs. Currently there is no way to opt-out the OpenTelemetry export.
```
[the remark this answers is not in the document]
roger2hk: That's the question I would ask after seeing this flag. I think it would be great if there is a way to opt-out when this is not required.

For running multiple instances, do you have more to share? Please feel free to add a suggestion.

docs/configuration.md
roger2hk: Added the checkpoint interval section with a Tessera limit table.

roger2hk: Let me create a separate pull request to update the default checkpoint interval value to 1500ms.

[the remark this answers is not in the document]
roger2hk: Here is the GCP minimum permitted checkpoint interval const.

https://github.com/transparency-dev/tessera/blob/main/storage/gcp/gcp.go#L69-L73

Here is the GCP check on the minimum permitted checkpoint interval.

https://github.com/transparency-dev/tessera/blob/main/storage/gcp/gcp.go#L230-L232

PR #382, opened 2025-06-25
Parallel and indempotent writes for issuers

phbnf:
Much inspired by Tessera's [gcp setObject](https://github.com/transparency-dev/tessera/blob/3fcd51d6bf6389f1c89b4f43858e09b194649d3d/storage/gcp/gcp.go#L1126), and [aws setObjectIfNoneMatch](https://github.com/transparency-dev/tessera/blob/3fcd51d6bf6389f1c89b4f43858e09b194649d3d/storage/aws/aws.go#L1362)

### DIFF
--- a/storage/aws/issuers.go
+++ b/storage/aws/issuers.go
@@ -19,13 +19,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"path"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/smithy-go"
+	"github.com/google/go-cmp/cmp"
 	"github.com/transparency-dev/tesseract/storage"
+	"golang.org/x/sync/errgroup"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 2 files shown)


storage/aws/issuers.go
```diff
@@ -62,31 +65,66 @@ func (s *IssuersStorage) keyToObjName(key []byte) string {
...
+				ContentType: aws.String(s.contentType),
+				IfNoneMatch: aws.String("*"),
+			}
+
+			// If we run into a precondition failure error, check that the object
+			// which exists contains the same content that we want to write.
+			// If so, we can consider this write to be idempotently successful.
+			if _, err := s.s3Client.PutObject(ctx, put); err != nil {
+				var apiErr smithy.APIError
+				if errors.As(err, &apiErr); apiErr.ErrorCode() == "PreconditionFailed" {
```
AlCutter, line 88: I think this is probably fine, but _technically_ should it be `if errors.As(err, &apiErr) && apiErr.ErrorCode() == ...` in order to not use uninitialised contents of `apiErr` if it's not the right sort of error?

In practice, unless "PreconditionFailed" is the default value for `ErrorCode()` on an uninitialised `APIError` it probably boils down to the same result - hitting the return on L101, but it triggered me nonetheless :)

storage/aws/issuers.go
```diff
@@ -62,31 +65,66 @@ func (s *IssuersStorage) keyToObjName(key []byte) string {
...
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(obj),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("getIssuer: failed to create reader for issuer %q in bucket %q: %w", obj, s.bucket, err)
+	}
+
+	d, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, fmt.Errorf("getIssuer: failed to read %q: %v", obj, err)
```
AlCutter, line 127: `r.Body.Close` won't be called if we hit here - is that ok?

storage/aws/issuers.go
```diff
@@ -62,31 +65,66 @@ func (s *IssuersStorage) keyToObjName(key []byte) string {
...
+			klog.V(2).Infof("AddIssuersIfNotExist: added %q in bucket %q", objName, s.bucket)
+			return nil
+		})
+	}
+
+	if err := errG.Wait(); err != nil {
+		return err
 	}
+
 	return nil
```
AlCutter, line 112: Can just make this `return errG.Wait()` since we don't wrap the error.

storage/gcp/issuers.go
```diff
@@ -57,35 +61,62 @@ func (s *IssuersStorage) keyToObjName(key []byte) string {
...
-		klog.V(2).Infof("AddIssuersIfNotExist: added %q in bucket %q", objName, s.bucket.BucketName())
+	if err := errG.Wait(); err != nil {
+		return err
 	}
+
 	return nil
 }
+
+// getIssuer returns the data and generation of the specified issuer, or an error.
+func (s *IssuersStorage) getIssuer(ctx context.Context, key string) ([]byte, int64, error) {
```
AlCutter, line 111: Do you need to return the object generation from here?
I think you're only ever going to have a single generation, and aren't going to try to do anything like "overwrite iff object is generation X"?

AlCutter (approved):

phbnf:
This PR is old and I didn't take the time to get back to it. It's not very important, so closing for now.

PR #383, opened 2025-06-25
Add outline of static-ct fsck tool

AlCutter:
This PR adds a first-cut outline of a correctness/self-integrity checking tool (`fsck`) for C2SP `static-ct` compliant logs.

TODO: needs transparency-dev/tessera#702 to be merged and a `go get`

Follow-up PRs could add functionality to verify that the set of intermediates required by the entries seen in the log are all present.


cmd/fsck/main.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/transparency-dev/tessera/fsck"
+	"github.com/transparency-dev/tesseract/internal/client"
+	"golang.org/x/crypto/cryptobyte"
+	"golang.org/x/mod/sumdb/note"
+	"golang.org/x/sync/errgroup"
+	"k8s.io/klog/v2"
+)
+
+var (
+	storageURL  = flag.String("storage_url", "", "Base tlog-tiles URL")
```
phbnf, line 42: nit, maybe rename to "monitoring_url" for consistency with the specs?

cmd/fsck/main.go
```diff
@@ -0,0 +1,266 @@
...
+	src, err := client.NewHTTPFetcher(logURL, hc)
+	if err != nil {
+		klog.Exitf("Failed to create HTTP fetcher: %v", err)
+	}
+	if *bearerToken != "" {
+		src.SetAuthorizationHeader(fmt.Sprintf("Bearer %s", *bearerToken))
+	}
+	v := verifierFromFlags()
+	if *origin == "" {
+		*origin = v.Name()
```
phbnf, line 74: I think you can remove this check? `verifierFromFlags` will fail if origin is empty.

cmd/fsck/main.go
```diff
@@ -0,0 +1,266 @@
...
+
+	klog.Info("OK")
+}
+
+func checkIntermediates(ctx context.Context, lsc *logStateCollector, readIssuer func(context.Context, []byte) ([]byte, error)) error {
+	klog.Infof("Checking intermediates CAS")
+	n := 0
+	work := make(chan []byte, 10)
+	eg := errgroup.Group{}
+	for range 10 {
```
phbnf, line 93: Should this be configurable? Or maybe set to the same value (or half?) that N is at?

cmd/fsck/main.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// merkleLeafHasher returns a function which knows how to:
+// - calculate RFC6962 Merkle leaf hashes for entries in a Static-CT formatted entry bundle,
+// - keep track of the set of intermediate cert fingerprints seen while parsing entry bundles.
+func (l *logStateCollector) merkleLeafHasher() func(bundle []byte) ([][]byte, error) {
+	return func(bundle []byte) ([][]byte, error) {
+		r := make([][]byte, 0, layout.EntryBundleWidth)
+		b := cryptobyte.String(bundle)
+		for i := 0; i < layout.EntryBundleWidth && !b.Empty(); i++ {
```
phbnf, line 142: I think you should be able to use something like https://github.com/transparency-dev/tesseract/blob/024f1e989970d41505590d1c65c3b3e7ed9cba34/internal/types/staticct/staticct.go#L47 or https://github.com/transparency-dev/tesseract/blob/024f1e989970d41505590d1c65c3b3e7ed9cba34/internal/types/staticct/staticct.go#L191 It might not work out of the box, but that's working as intended: I'd like to understand what we need these parsing functions for exactly before having one to rule them all. If you can't find an easy way to use any of these methods, put the parsing method you need over there, and at least they will all be in the same file.

Now, here's an easy counter argument: this is a check tool..... so maybe it *should* reimplement everything?

cmd/fsck/main.go
AlCutter: Good spot, ta!

cmd/fsck/main.go
AlCutter: The assumption was that there weren't going to be (relative to the number of tiles/entrybundles just fetched) very many issuers, so some small amount of parallelism was "good enough", making it `N` is fine too, though - done!

phbnf (approved):

PR #402, opened 2025-06-30
Update fetcher contract & impl to fallback to full resources

AlCutter:
This PR provides GC-compliant operation to all users of the `internal/client` package by pushing the responsibility of falling-back to the corresponding full resource when a partial resource is not found.

Analogous to transparency-dev/tessera#704.


AlCutter:
> Doe this make #398 obsolete?

Yeah, they'll both do it similar ways now - I'll close the other one.

PR #408, opened 2025-07-09
Fetcher checks content type

phbnf:
Matching PR incoming to tessera if you're happy with it!

This PR checks that Content-Type is exactly of length 1, and that it has the content type specified in specs. We could be a bit more lax and use `slices.Contains`, especially since the Content-Type header could be set multiple times if I understand correctly, but this would deviate a bit from specs.


phbnf:
We spoke about this a while back and agreed that this should not be inside the client. Closing.

PR #444, opened 2025-07-16
Replace `Terraform` with `OpenTofu` in docs or comments

roger2hk:
Towards https://github.com/transparency-dev/tesseract/issues/307.


deployment/live/aws/conformance/ci/README.md
```diff
@@ -36,6 +36,6 @@ export TESSERACT_PREFIX_NAME={VALUE}
 Resources will be named using a `${TESSERACT_PREFIX_NAME}-${TESSERACT_BASE_NAME}`
 convention.
 
-Terraforming the project can be done by:
+OpenTofuing the project can be done by:
```
mhutchinson, line 39: ...is that the verb?!

mhutchinson: Interested in @AlCutter thoughts here. The slash form might work, because ~everyone already knows what "terraforming" is, and then it naturally implies what the second term means.

phbnf: +1, it looks a bit odd. "Deploy the Terraform config with OpenTofu:". Also, maybe in a different PR, I think we should avoid telling folks to change directories, and rather use: `terragrunt apply --terragrunt-working-dir=<PATH HERE>`

roger2hk: Updated "Deploy the Terraform config with OpenTofu:". I will update to use `terragrunt apply --terragrunt-working-dir=<PATH HERE>` in another pull request.

Issue #322, opened 2025-05-21
Cache SCT timestamps


phbnf:
Now that deduplication has been migrated to Tessera's deduplication, each time an index is returned, TesseraCT reads the leaf from GCS/S3 and extract the timestamp out of it and build the SCT again. We can cache this timestamp to make subsequent similar requests faster. We could also consider:
 - caching the entire SCT (this would have parsing implications, as well as memory usage impact)
 - caching the timestamp and the index, then disabling antispam caching on Tessera (this would cut memory usage and remove layered caches)
 - caching the timestamp of all entries in the bundle from which an entry is being deduped for a given request (To get the timestamp of entry X in a bundle, all the entries with a smaller index need to be parsed first, so their timestamp would be available.  This might not be a good idea depending on user query patterns though.)

PR #445, opened 2025-07-17
Limit the number of duplicate requests in flight

phbnf:
Towards #322, #448

### DIFF
--- a/cmd/experimental/posix/main.go
+++ b/cmd/experimental/posix/main.go
@@ -59,6 +59,7 @@ var (
 	origin                   = flag.String("origin", "", "Origin of the log, for checkpoints and the monitoring prefix.")
 	storageDir               = flag.String("storage_dir", "", "Path to root of log storage.")
 	pushbackMaxOutstanding   = flag.Uint("pushback_max_outstanding", tessera.DefaultPushbackMaxOutstanding, "Maximum number of number of in-flight add requests - i.e. the number of entries with sequence numbers assigned, but which are not yet integrated into the log.")
+	pushbackMaxDupesInFlight = flag.Uint("pushback_max_dupes_in_flight", 1024, "Maximum number of number of in-flight duplicate add requests. When 0, duplicate entries are always pushed back.")
 	rootsPemFile             = flag.String("roots_pem_file", "", "Path to the file containing root certificates that are acceptable to the log. The certs are served through get-roots endpoint.")
 	rejectExpired            = flag.Bool("reject_expired", false, "If true then the certificate validity period will be checked against the current time during the validation of submissions. This will cause expired certificates to be rejected.")
 	rejectUnexpired          = flag.Bool("reject_unexpired", false, "If true then CTFE rejects certificates that are either currently valid or not yet valid.")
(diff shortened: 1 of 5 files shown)


cmd/experimental/posix/main.go
```diff
@@ -168,7 +169,7 @@ func newStorage(ctx context.Context, signer note.Signer) (*storage.CTStorage, er
 		return nil, fmt.Errorf("failed to initialize GCP issuer storage: %v", err)
 	}
 
-	return storage.NewCTStorage(ctx, appender, issuerStorage, reader, *enablePublicationAwaiter)
+	return storage.NewCTStorage(ctx, appender, issuerStorage, reader, *enablePublicationAwaiter, *pushbackMaxDupesInFlight)
```
AlCutter, line 172: Feels like this is getting close to warranting a `CTStorageOptions` struct

storage/storage.go
```diff
@@ -138,6 +155,10 @@ func (cts *CTStorage) Add(ctx context.Context, entry *ctonly.Entry) (uint64, uin
 		}
 	}
 	if idx.IsDup {
+		if cts.dupesInFlight.Load() > int64(cts.maxDupesInFlight) {
+			return 0, 0, fmt.Errorf("antispam in-flight %w", tessera.ErrPushback)
```
AlCutter, line 159:
```suggestion
			return 0, 0, fmt.Errorf("too many duplicate submissions %w", tessera.ErrPushback)
```

AlCutter (approved):
Nice one, just a couple of little suggestions in-line.

cmd/experimental/posix/main.go
roger2hk: +1 (it's getting longer and longer)

cmd/experimental/posix/main.go
```diff
@@ -59,6 +59,7 @@ var (
 	origin                   = flag.String("origin", "", "Origin of the log, for checkpoints and the monitoring prefix.")
 	storageDir               = flag.String("storage_dir", "", "Path to root of log storage.")
 	pushbackMaxOutstanding   = flag.Uint("pushback_max_outstanding", tessera.DefaultPushbackMaxOutstanding, "Maximum number of number of in-flight add requests - i.e. the number of entries with sequence numbers assigned, but which are not yet integrated into the log.")
+	pushbackMaxDupesInFlight = flag.Uint("pushback_max_dupes_in_flight", 100, "Maximum number of number of in-flight duplicate add requests. When 0, duplicate entries are always pushed back.")
```
roger2hk, line 62: I think we can remove `_in_flight` from the flag name. The `pushback_max_outstanding` flag is also for in-flight requests and we don't have "in flight" in the flag name.

storage/storage.go
```diff
@@ -57,23 +58,39 @@ type IssuerStorage interface {
...
-	reader        tessera.LogReader
-	awaiter       *tessera.PublicationAwaiter
-	enableAwaiter bool
+	storeData        func(context.Context, *ctonly.Entry) tessera.IndexFuture
+	storeIssuers     func(context.Context, []KV) error
+	reader           tessera.LogReader
+	awaiter          *tessera.PublicationAwaiter
+	enableAwaiter    bool
+	dupesInFlight    atomic.Int64
+	maxDupesInFlight uint
```
roger2hk, line 67:
```suggestion
	maxDupesInFlight uint
	
	dupesInFlight atomic.Int64
```

Separating the struct variables into two different groups helps the readability.

roger2hk (approved):

cmd/experimental/posix/main.go
phbnf: Done. We now have to pass a struct to a function that itself create a new struct... We should be able to restructure the code a bit later, but let's leave it like this for now to see what we'll need to add / remove.

AlCutter (approved):

storage/storage.go
phbnf: I agree that we could probably re-arrange this struct. I can't really tell what grouping you were thinking of and why dupesInFlight should be alone though. I'll merge this as is, and we can talk about it when you're back :)

PR #456, opened 2025-07-22
Avoid JSON roundtrip

AlCutter:
This gives a 60% reduction in allocs & CPU usage in `LogWriter.Run()`.

Before:
```
BenchmarkLogWriter-30               7486           1601835 ns/op          474228 B/op      17013 allocs/op
```

After:
```
BenchmarkLogWriter-30              20853            574608 ns/op          137705 B/op       7010 allocs/op
```


AlCutter:
Is this helpful, or is it pushing in the wrong direction?

For context, the intent was to try to streamline the hammer a bit - at high QPS testing it's using far more CPU than the TesseraCT server under test does.

phbnf:
Fine by me - it takes the hammer further away from the Tessera one, but I think this is working as intended: we need to see what modifications are required before/ifWeEver re-unite them.

Just curious - do you have any performance number by any chance?

AlCutter:
> Just curious - do you have any performance number by any chance?

Added a rough-and-ready benchmark.

PR #461, opened 2025-07-23
Remove prefix

phbnf:
Towards #104 

According to https://c2sp.org/static-ct-api, `and the origin line MUST be the submission prefix of the log as a schema-less URL with no trailing slashes`.

There are two parts to a submission prefix: the host, and the path prefix. Before this PR, TesseraCT ignored the host, and required for the path to be prefixed with the origin: it was not possible to bring up a specs compliant log. 

This PR removes the path prefix requirement, and makes it possible to bring up a specs compliant log behind https://HOST/c1/v1, where HOST = origin. Removing all requirements from TesseraCT is the easiest approach. The only downside I can see, is that for local and non-public deployments, the origin is not in the URL anymore, which is quite a handy thing to have.

Alternatively/After, we could:
 - configure _any_ prefix via flag: this would allow A. keeping the origin in the URL for local deployments B. origins split between a host and the path like: "arche.staging.ct.transparency.dev/2025h1". 
 - Make TesseraCT check the submission prefix across the host path for every request it receives: this wouldn't be bulletproof since proxies can still fiddle URLs and co, but would at least make TesseraCT somewhat opinionated.

### DIFF
--- a/cmd/experimental/posix/README.md
+++ b/cmd/experimental/posix/README.md
@@ -9,7 +9,7 @@ production log!
 Generate an ECDSA key like so:
 
 ```bash
-$ openssl ecparam -name prime256v1 -genkey -noout -out test-ecdsa-priv.pem 
+openssl ecparam -name prime256v1 -genkey -noout -out test-ecdsa-priv.pem 
 ```
 
 And then start a log with the following command:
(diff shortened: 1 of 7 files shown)


cmd/experimental/posix/README.md
```diff
@@ -33,7 +33,7 @@ You can try "preloading" the log with the contents of another CT log, e.g.:
 
 ```bash
 go run github.com/google/certificate-transparency-go/preload/preloader@master \
-  --target_log_uri=http://localhost:6962/example.com/test-ecdsa \
+  --target_log_uri=http://localhost:6962/\
```
roger2hk, line 36:
```suggestion
  --target_log_uri=http://localhost:6962/ \
```

phbnf:
Candidate other implementation: https://github.com/transparency-dev/tesseract/pull/462

phbnf:
We've gone with #462 instead.

PR #463, opened 2025-07-23
Move POSIX binary into /cmd/tesseract

AlCutter:
This PR moves the POSIX `tesseract` binary into the same location as the others, and updates the README to provide a bit more guidance around suitable filesystems.


cmd/tesseract/posix/README.md
phbnf, line 1: A few nits:

 - the tesseract command needs to be updated to remove the "experimental" bit
 - Remove the output of the tesseract command or move it in a different snippet to allow folks to copy/paste it.
 - It uses arche2025h1's roots, but preloads from xenon2025h2. It's actually because these roots match... but we can make it a bit neater. There are further instructions in `main/deployment/live/gcp/test` and in `main/deployment/live/aws/test`
 - I've tried to brand the getting started bits of doc as "Codelab" to be catchy. Might be worth doing the same thing here for consistency, and then to edit the codelab sections in `/docs/README.md` and `/README.md`.
 - We also need to edit `/cmd/tesseract/README.md` (or find a new home for all the Dockerfiles and co)
 - I don't think that we need add config files under `/deployment` for POSIX, it doesn't really make sense. But a pointer from `/deployment/README.md` to `/cmd/posix/README.md` could be useful?

cmd/tesseract/posix/README.md
```diff
@@ -1,8 +1,21 @@
 # POSIX static-ct server
 
-This directory contains an experimental `static-ct` server which uses Tessera's
-POSIX backend for storing the log.  It's not yet ready to be used as a
-production log!
+This directory contains a `static-ct` server which uses
```
phbnf, line 3: `tesseract` server? If this is a recent copy paste from somewhere else in the repo, let's get this fixed as well!

phbnf (approved):
I've also filed https://github.com/transparency-dev/tesseract/issues/465, which I think should be a requirement for anything outside of experimental to match other implementations.

Also, should we add metrics to the binary? IMO, one of the first things folks should do when bringing something up, is to look at monitoring!

AlCutter:
> I've also filed #465, which I think should be a requirement for anything outside of experimental to match other implementations.
> 
> Also, should we add metrics to the binary? IMO, one of the first things folks should do when bringing something up, is to look at monitoring!

Ta - 100% agree, these will come in follow-ups.

cmd/tesseract/posix/README.md
AlCutter: This meant `static-ct` as in a server which complies with the `static-ct` spec.

cmd/tesseract/posix/README.md
AlCutter: Great spots, ta!
done.

PR #470, opened 2025-07-25
Add some POSIX performance numbers

AlCutter:
Add some performance numbers for the POSIX binary.


phbnf:
@roger2hk for consistency - it would be great to list flags on other load tests?

phbnf (approved):

roger2hk:
> @roger2hk for consistency - it would be great to list flags on other load tests?

Sounds good. The default flag values were used in most of the cases.

PR #478, opened 2025-07-28
Add local signer as alternative to AWS Secrets Manager

pimvanpelt:
The AWS personality makes an assumtion of RDS (can be replaced by MySQL), S3 (can be replaced by a local S3-compatible endpoint), and Secret Manager (hardcoded). This PR allows for an operator to add a local key signer module based off the local filesystem, by specifying `-signerPublicKeyFile` and `-signerPrivateKeyFile` flags, which are mutually exclusive with the `-*ScretName` flags.

It allows the AWS personality to  run entirely on-prem.

See [blogpost](https://ipng.ch/s/articles/2025/07/26/certificate-transparency-part-1/) for a paper on self-hosted S3/MySQL versus POSIX.

### DIFF
--- a/cmd/tesseract/aws/local_signer.go
+++ b/cmd/tesseract/aws/local_signer.go
@@ -1,0 +1,94 @@
+// Copyright 2025 The Tessera authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


phbnf:
Adding @roger2hk since he's been working on the Secret Manager implementation.

Looks good to me, thank you for the PR. For the record, we might change things a bit later as we work on #219. Until then, we're trying to keep key configuration as minimal as possible, and this is pretty much what this PR does :)

roger2hk:
Thanks for the contribution. 😄

PR #494, opened 2025-08-05
S3+MySQL load test numbers

phbnf:
Towards #321 

While I'm there, format this file.

Spoke with [pimvanpelt](https://github.com/pimvanpelt) before who agreed for us to mention these numbers.

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -379,15 +379,15 @@ love to hear about them: [get in touch](../README.md#wave-contact)!
 #### MinIO MariaDB (from IPng Networks)
 
 This test was performed by [IPng Networks](https://ipng.ch/), around 2025-07-26.
-More details are available on their [blog post](https://ipng.ch/s/articles/2025/07/26/certificate-transparency-part-1/).
+More details are available on their [blog post](https://ipng.ch/s/articles/2025/07/26/certificate-transparency-part-1-tesseract/).
 Kudos to IPng and Pim!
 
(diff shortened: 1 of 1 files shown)


docs/performance.md
```diff
@@ -368,3 +369,30 @@ MiB Swap:      0.0 total,      0.0 free,      0.0 used.  30354.7 avail Mem
...
+SAS controller with 6pcs of 1.92TB enterprise storage (Samsung part
+number P1633N19).
+
+The log, hammer, MinIO and MariaDB were running on the same machine.
+
+TesseraCT sustained 500 write qps for a few hours, with:
+
+* TesseraCT using about 2.9 CPUs/s
+* MariaDB using 0.3 CPUs/s
+* The hammer using 6.0 CPUs/s
```
mhutchinson, line 396: We can't change the published numbers, but this number is already likely improved since #456.

pimvanpelt:
PTAL one note on L390.

docs/performance.md
phbnf: Good point, I added some indication about when these test were done.

phbnf:
@pimvanpelt I presume that I got something wrong on line 390? I had another careful read at the article, is my understanding correct that TesseraCT, the Hammer and MariaDB were running on `ctlog-test`, while MinIO was running on a separate `minio-ssd` machine? I can tell from the last screenshot that `ctlog-test` was also running with one (two?) `e5-2640` processor(s)?

PR #510, opened 2025-08-13
Improve AWS codelab doc

roger2hk:


phbnf:
Not sure if you've seen my comment above: I think the instructions to connect the database should be outside of the note since these instructions should always be followed. It's not a big dealbreaker, I'll approve it in the interest of moving forward.

roger2hk:
> Not sure if you've seen my comment above: I think the instructions to connect the database should be outside of the note since these instructions should always be followed. It's not a big dealbreaker, I'll approve it in the interest of moving forward.

Oh... I thought the suggested change was to add some extra lines. I think it's more reasonable to keep the 1st run section separate.

PR #531, opened 2025-08-15
Add base name in aws codelab

phbnf:
I think that it's missing? Towards #321


AlCutter (approved):

roger2hk (approved):

roger2hk:
That's optional but you can override it. It may be overkilled in a codelab.

PR #563, opened 2025-09-18
Add support for Tink keyset signer

Hayden-IO:
This PR adds support for in-memory signing using a Tink keyset. The keyset is encrypted with a key-encryption-key stored in GCP KMS. The key is decrypted on startup and loaded into memory. This uses a utility to unpack the keyset into a crypto.Signer so that it can be used to sign certificates. This also validates that the key is an ECDSA P-256 key as per RFC 6962, since Tink supports many key types.

### DIFF
--- a/cmd/tesseract/gcp/main.go
+++ b/cmd/tesseract/gcp/main.go
@@ -17,6 +17,7 @@ package main
 
 import (
 	"context"
+	"crypto"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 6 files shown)


haydentherapper:
Reopening, ref to https://github.com/transparency-dev/tesseract/pull/282 for the previous discussion.

haydentherapper:
Bumping this, is there any interest in supporting this at this time?

Issue #538, opened 2025-08-26
[Demand Signal] Limit submission of old certificates


phbnf:
This would limit the submission of certificates that are old enough that they are not on the CA certificate issuance path. 

Much inspired by https://github.com/google/certificate-transparency-go/pull/1698. This was introduced to protect log resources, and make sure they would be spent on critical submissions. TesseraCT should be performant enough that this is not required, but opening this just in case.

mhutchinson:
This seems like it would be easy enough to add? Configuration via a flag would allow this to be enabled/disabled, and made more or less strict.

Where I particularly see this being useful is where a log operator has brought up a shard for a time period a bit late, and there are already tonnes of certs that the log could accept. The priority should always be given to fresh certs, but without this, there could be billions of other certs being forced into the log.

PR #570, opened 2025-09-24
Support rate limiting of old submissions

AlCutter:
This PR adds support for an (optional) simple rate limiter which can be used to cheaply throttle submissions of certs older than a given threshold.

Towards #538


phbnf (approved):

AlCutter:
> As we spoke about yesterday, it would be great to modify method APIs such that we can apply the rate limit after parsing the first certificate only, and not the full chain.

Agreed, although I think that probably merits being its own follow-up PR to avoid blowing this on up in size.

PR #576, opened 2025-09-29
Add notBefore age histogram

AlCutter:
This PR adds a histogram metric for inspecting the `notBefore` age of submissions.

### DIFF
--- a/internal/ct/handlers.go
+++ b/internal/ct/handlers.go
@@ -64,15 +64,15 @@ const (
 var (
 	// Metrics are all per-log (label "origin"), but may also be
 	// per-entrypoint (label "ep") or per-return-code (label "rc").
-	once                sync.Once
-	knownLogs           metric.Int64Gauge       // origin => value (always 1.0)
-	lastSCTIndex        metric.Int64Gauge       // origin => value
-	lastSCTTimestamp    metric.Int64Gauge       // origin => value
-	reqCounter          metric.Int64Counter     // origin, op => value
-	rspCounter          metric.Int64Counter     // origin, op, code => value
-	reqDuration         metric.Float64Histogram // origin, op, code => value
-	rateLimitedRequests metric.Int64Counter     // origin, reason
-	notBeforeAge        metric.Float64Histogram // origin ==> value
+	once                   sync.Once
+	knownLogs              metric.Int64Gauge       // origin => value (always 1.0)
+	lastSCTIndex           metric.Int64Gauge       // origin => value
+	lastSCTTimestamp       metric.Int64Gauge       // origin => value
+	reqCounter             metric.Int64Counter     // origin, op => value
+	rspCounter             metric.Int64Counter     // origin, op, code => value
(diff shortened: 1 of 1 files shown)


internal/ct/handlers.go
```diff
@@ -319,6 +325,8 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
 		return http.StatusBadRequest, nil, fmt.Errorf("failed to parse add-chain contents: %s", err)
 	}
 
+	notBeforeAge.Record(ctx, time.Since(chain[0].NotBefore).Seconds())
```
phbnf, line 328: Given that this happens before we verify the chain - on purpose since this information will be useful to have insights into rate limiting, this won't provide information about what actually gets into the log, which could be misleading. How about we either:
 1. add an "accepted / rate_limited" field
 2. tag this metric with something like `tesseract.notbefore.age.unverified`

I think I have a slight preference for 1, but I'm keen to hear what you think!

internal/otel/cast.go
```diff
@@ -14,14 +14,38 @@
...
 	// We use seconds as units, Open Telemetry's default unit for time.
-	SubSecondLatencyHistogramBuckets = []float64{0, 10e-3, 50e-3, 100e-3, 200e-3, 300e-3, 400e-3, 500e-3, 600e-3, 700e-3, 800e-3, 900e-3, 1000e-3, 1200e-3, 1400e-3, 1600e-3, 1800e-3, 2000e-3, 2500e-3, 3000e-3, 4000e-3, 5000e-3, 6000e-3, 8000e-3, 10000e-3}
+	SubSecondLatencyHistogramBuckets = createBuckets(
+		[]float64{0, 10, 50, 100, 200, 300, 400, 500, 600, 700, 800, 900, 1000, 1200, 1400, 1600, 1800, 2000, 2500, 3000, 4000, 5000, 6000, 8000, 10000},
+		time.Millisecond,
+		time.Millisecond)
+
+	// SubmissionAgeHistogramBuckets is an exponential range of buckets defined in seconds and intended to cover some reasonable ages of submissions.
+	// This range is based on the default OTel buckets, but ranges from ~5s at the low end up to ~70days.
+	SubmissionAgeHistogramBuckets = createBuckets(
```
phbnf, line 32: I think that each number defined an excluded upper boundary, right? Do you think we could align buckets with common time boundaries: 24h, 48h, 1w, 1month, etc? Right now it's mostly aligned with it, but not totally. Our logs are configured to rate limit after 28h, so we should at least either modify this value / align the metric with this?

AlCutter: I can do that, but we then end up with irregular bucket widths - I read _some stuff_ which advised against doing that, so I tried to get something close but which stuck to the norms OTel has gone with.

On the 28h front, I figured that this is likely to change, and quite possibly as a result of the PR so it was ok that they were not bound in some way.

internal/ct/handlers.go
phbnf: (2) LGTM!

Issue #555, opened 2025-09-10
Request counters for 429 could tell what's driving the 429


phbnf:
429 can be driven by:
 - too many duplicates in flights
 - too big of an antispam lag
 - too big of a sequencing queue

It would be great for TesseraCT request counters to tell the difference between them. `tessera.appender.calls` already makes the distinction between them, but not `tesseract.http` metrics.

PR #577, opened 2025-09-29
Better 429 and rate limit reasons

phbnf:
Towards #555, #538.

This PR:
 - changes the rate limit reason from "too many old certs" to "old_cert" to remove spaces and be more concise
 - adds a reason to 429s, with a "rate_limit_" prefix if it's a rate limit, "tessera_pushback_" if it's a tessera pushback!
 
 I'll send more rate limits metric edits in followup PRs, for duplicates, and issuers.


internal/ct/handlers.go
```diff
@@ -351,13 +351,13 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
 	index, dedupTimeMillis, err := log.storage.Add(ctx, entry)
 	if err != nil {
 		if errors.Is(err, tessera.ErrPushback) {
+			tooManyRequestsAttribute := fmt.Sprintf("tessera_pushback_%s", strings.Split(err.Error(), " ")[0])
```
AlCutter, line 354: This adds a monitoring-critical dep on the _text_ of the returned error, is there a way we can make this safer?

AlCutter, line 354: Also, it feels a bit cheeky to be adding a `tessera_` scoped attribute here :) (maybe that's fine, though?)

phbnf: We can add a field to pushback errors in Tessera if you're ok with this?

AlCutter: Yeah, I think that could work...

PR #597, opened 2025-10-08
Update lockfiles

AlCutter:
From applying #595


deployment/live/gcp/static-ct-staging/logs/arche2025h1/.terraform.lock.hcl
```diff
@@ -1,55 +1,61 @@
-# This file is maintained automatically by "tofu init".
+# This file is maintained automatically by "terraform init".
```
roger2hk, line 1: Try again using `tofu init`?

roger2hk (approved):

PR #607, opened 2025-10-16
Fix issuers bug

### LINKED ISSUE #606: Missing issuers in GCP and AWS
This is likely caused by [this return statement](https://github.com/transparency-dev/tesseract/blob/42e6d0c6d93dbd1a3941902f33b0bea978850185/storage/gcp/issuers.go#L110), which really, should be a continue.

It's only impacting AWS and GCP, not POSIX.

Let's fix one of the TODO an use error groups + parallel writes while we're there.

AlCutter:
This PR fixes a couple of issues in the GCP and AWS issuer storage implementations:
- In certain conditions, it's possible that intermediates could be lost without any error being surfaced.
- A SEGV if an error is returned from the S3 client which isn't an instance of `smithy.APIError`.

Fixes #606

### DIFF
--- a/storage/aws/issuers_test.go
+++ b/storage/aws/issuers_test.go
@@ -148,17 +148,6 @@ func TestAddIssuersIfNotExist(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name: "add existing issuer with different data",
-			kv: []storage.KV{
-				{K: []byte("issuer1"), V: []byte("issuer1 data")},
-				{K: []byte("issuer1"), V: []byte("different data")},
-			},
-			want: map[string][]byte{
-				"issuer1": []byte("issuer1 data"),
-			},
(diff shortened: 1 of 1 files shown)


storage/aws/issuers.go
```diff
@@ -83,6 +86,8 @@ func (s *IssuersStorage) keyToObjName(key []byte) string {
 
 // AddIssuers stores Issuers values under their Key if there isn't an object under Key already.
 func (s *IssuersStorage) AddIssuersIfNotExist(ctx context.Context, kv []storage.KV) error {
+	eg := errgroup.Group{}
+
 	// We first try and see if this issuer cert has already been stored since reads
```
phbnf: I think this is outdated or misplaced

phbnf (approved):

PR #608, opened 2025-10-17
Add remora witness to staging policy

AlCutter:
Adds a 3rd party dev witness to the staging policy: https://remora.n621.de/


phbnf:
A new word for me.

AlCutter:
[This is now applied to all 3 arche logs]

PR #622, opened 2025-10-29
More orgin bells and whistles 

phbnf:
This should help log operators to not misconfigure the origin.

A warning will be shown if a request is received on a submission endpoint that does not match with the log's origin.
While I'm there, improve the origin flag description.

### DIFF
--- a/internal/ct/handlers.go
+++ b/internal/ct/handlers.go
@@ -155,30 +155,30 @@ func submissionEndpoint(r *http.Request) (string, error) {
 	return fmt.Sprintf("%s%s", host, r.URL.Path), nil
 }
 
-// receivedAtOrigin returns an empty string if r was received on a URL starting with origin.
-// If not, it returns the URL the request was received at, without the port and the query string.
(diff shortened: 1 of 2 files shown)


cmd/tesseract/posix/main.go
```diff
@@ -64,7 +64,7 @@ var (
 	// Functionality flags
 	httpEndpoint             = flag.String("http_endpoint", "localhost:6962", "Endpoint for HTTP (host:port).")
 	maskInternalErrors       = flag.Bool("mask_internal_errors", false, "Don't return error strings with Internal Server Error HTTP responses.")
-	origin                   = flag.String("origin", "", "Origin of the log, for checkpoints.")
+	origin                   = flag.String("origin", "", "Origin of the log, for checkpoints. This MUST match the log's submission prefix as per https://c2sp.org/static-ct-api.")
```
AlCutter: Do we already check that there is no trailing slash on the provided origin? If not, might be worth adding too?

AlCutter (approved):

Issue #644, opened 2025-12-04
TesseraCT partial tile cleanup


pimvanpelt:
**Description**

I archived IPng's `halloumi2026h2` logshard with 4765491 leaves, which was consuming 146GB on disk.
Using Geomys' `photocamera` app, turned the log into a 4.0GB zipfile. Unzipping that, this yields 4.4GB of data on disk. The resulting unzipped data passes `ct-fsck`:

```
pim@summer:/ssd-vol0/logs$ du -sch tile/ issuer/
4.4G    tile/
946K    issuer/
4.4G    total
pim@summer:/ssd-vol0/logs$ ~pim/bin/ct-fsck -origin halloumi2026h2.log.ct.ipng.ch         -monitoring_url file:///ssd-vol0/logs/         -public_key="MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAENDhVrTtcQebbk/R6YW8Gt1RW+WZ8qscHUsI2/TW5f7bu+mEst2lskXbR9HEBQvyhkYKVTuigQjgZvDYqRhcn0Q=="         -N 20 -user_agent_info="ct-ops@ipng.ch"
I1205 00:12:37.997709 2546513 main.go:335] Using verifier string: halloumi2026h2.log.ct.ipng.ch+53348351+BTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABDQ4Va07XEHm25P0emFvBrdUVvlmfKrHB1LCNv01uX+27vphLLdpbJF20fRxAUL8oZGClU7ooEI4Gbw2KkYXJ9E=
I1205 00:12:37.997827 2546513 main.go:112] Checking issuers CAS
I1205 00:12:37.998584 2546513 fsck.go:56] Fsck: checking log of size 4765491
I1205 00:12:54.667362 2546513 fsck.go:111] Successfully fsck'd log with size 4765491 and root ucxfVJxhzGDvsFzvHEmV/U+p3cvjWSHuwLhy+ljp17Q= (b9cc5f549c61cc60efb05cef1c4995fd4fa9ddcbe35921eec0b872fa58e9d7b4)
I1205 00:12:54.667397 2546513 main.go:82] OK
pim@summer:/ssd-vol0/logs$ cat checkpoint 
halloumi2026h2.log.ct.ipng.ch
4765491
ucxfVJxhzGDvsFzvHEmV/U+p3cvjWSHuwLhy+ljp17Q=

[...]

PR #649, opened 2025-12-09
Decrease default interval between garbage collection runs, and add documentation

phbnf:
Towards #644

### DIFF
--- a/cmd/tesseract/README.md
+++ b/cmd/tesseract/README.md
@@ -201,9 +201,17 @@ resources TesseraCT spends on servicing duplicate requests.
 
 The `garbage_collection_interval` flag controls Tessera's Garbage Collection.
 It controls how often Tessera scans for partial tiles and entry bundles, and
-deletes them if a corresponding full tile or entry bundle has been published.
-Each storage implementation keeps track of the highest entry index that was
-garbage collected. Every `garbage_collection_interval`, Tessera pre-computes
+deletes them if a corresponding full tile or entry bundle has since been
+published. Tessera automatically keeps track of the garbage collection progress
+through the tree, keeping up with new entries but avoiding re-examining earlier
+parts which have already been processed. To achieve this, it examines the log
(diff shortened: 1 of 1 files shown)


cmd/tesseract/README.md
```diff
@@ -201,7 +201,21 @@ resources TesseraCT spends on servicing duplicate requests.
...
 The `garbage_collection_interval` flag controls Tessera's Garbage Collection.
 It controls how often Tessera scans for partial tiles and entry bundles, and
-deletes them if a matching full tile or entry bundle has been published.
+deletes them if a matching full tile or entry bundle has been published. Each
+storage implementation keeps track of the highest entry index that was garbage
+collected. Every `garbage_collection_interval`, Tessera pre-computes the potential
+paths of [100](https://github.com/search?q=repo%3Atransparency-dev%2Ftessera+maxBundlesPerRun&type=code)
+partial entry bundle and partial tile directories. It then attempts to delete them
+one after another. If a path resolves to an existing partial entry bundle or tile
+directory, it is deleted. Otherwise, nothing happens and it moves to the next one.
```
AlCutter: This is a bit focused on POSIX, maybe a more general description could be something like:

_Tessera examines the log sequentially in chunks, one chunk per `garbage_collection_interval`, removing any obsolete partial resources found within the current chunk. Currently, each chunk is defined as up-to 100 entry bundles along with the vertical "slice" of the merkle tree which covers these bundles._

cmd/tesseract/README.md
```diff
@@ -201,7 +201,21 @@ resources TesseraCT spends on servicing duplicate requests.
...
+
+To ensure that garbage collection is cleaning the log faster than it grows, 
+`100*256/garbage_collection_interval` must be higher than the log's throughput.
+
+If garbage collection was turned off, and is then turned on, it will take
```
AlCutter:
```suggestion
If garbage collection was disabled, and is later re-enabled, it will take
```

AlCutter (approved):

Issue #212, opened 2025-03-24
Root update process


phbnf:
To update `arche2025h1` roots, we fetch them from Google production logs, and filter out the ones we can't parse with the "new" x509 library. Let's uncouple this from Google's logs and implement a dedicated root update mechanism. https://github.com/sectigo/CTLogs-AcceptedRoots could be useful.

PR #657, opened 2025-12-15
Automatically fetch roots from ccadb

phbnf:
Towards #212 

This PR makes it possible for TesseraCT to fetch roots from CCADB.

This PR does two things:
 - makes it possible to update roots on the fly
 - calls the ccadb library at startup and on a clock to fetch new roots

A followup PR will make it possible to save these roots.

### DIFF
--- a/internal/x509util/pem_cert_pool.go
+++ b/internal/x509util/pem_cert_pool.go
@@ -44,7 +44,10 @@ type PEMCertPool struct {
 
 // NewPEMCertPool creates a new, empty, instance of PEMCertPool.
 func NewPEMCertPool() *PEMCertPool {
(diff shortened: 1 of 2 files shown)


internal/x509util/pem_cert_pool.go
```diff
@@ -42,52 +44,76 @@ type PEMCertPool struct {
 
 // NewPEMCertPool creates a new, empty, instance of PEMCertPool.
 func NewPEMCertPool() *PEMCertPool {
-	return &PEMCertPool{fingerprintToCertMap: make(map[[sha256.Size]byte]x509.Certificate), certPool: lax509.NewCertPool()}
+	return &PEMCertPool{mu: sync.RWMutex{}, fingerprintToCertMap: make(map[[sha256.Size]byte]x509.Certificate), certPool: lax509.NewCertPool()}
```
AlCutter: nit: probably worth making this multi-line now, it's quite long.

PR #678, opened 2026-01-22
GCP: Update codelab and terra{form,grunt} to use generate_key

AlCutter:
This PR externalizes the creation of log keys for GCP logs to the log operator, before turning up the log.

This removes the inherent danger of secret key leakage from relying on terraform to do it.

Keys are assumed, by default, to be named according to the scheme used by the `generate_key` tool, although an escape hatch is provided so that "legacy" GCP logs can still be supported.

Towards #219


phbnf:
As discussed online earlier today, let's get back to this tomorrow and use:
 - secret resource names directly, such that robots applying terragrunt won't need to access Secret Manager anymore
 - see if we can use lifecycle blocks to prevent the unfortunate deletion of old secrets

PR #684, opened 2026-01-26
Improve and fix GCP CI test

AlCutter:
This PR makes a few changes to the CI set up to fix the breakage and introduce some minor improvements:
- Correct naming of keys generated by the `generate_key` command. Previously, incorrectly `log` preceded the origin.
- Replace `origin_suffix` variable with `origin`. Previously, `origin` was passed implicitly as `base_name`, and `origin_suffix` was appended in all locations where an origin was required. This change makes it easier to be explicit and consistent about the value of origin across multiple locations.
- Derive a `safe_origin` internal string for use in naming GCP resources where only alphanumeric and `-` chars are permitted.
- Replace key suffix vars with the full secret manager resource ID for the CI log.
- Fix Cloud Build CI to work with the new assumptions about ownership of keys.
   + Keys for are now provisioned and deleted for each run by explicit steps which ensures that the `generate_key` tool is tested.
   + Update to specify `origin` and use that consistently.

### DIFF
--- a/deployment/modules/gcp/cloudbuild/conformance/main.tf
+++ b/deployment/modules/gcp/cloudbuild/conformance/main.tf
@@ -74,7 +74,7 @@ resource "google_cloudbuild_trigger" "build_trigger" {
       ]
     }
 
-    ## Destroy any pre-existing infrastructure.
+    ## Clean-up: Destroy any pre-existing infrastructure.
     ## This might happen if a previous cloud build failed for some reason.
     step {
       id     = "preclean_env"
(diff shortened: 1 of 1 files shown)


deployment/live/gcp/static-ct-staging/logs/arche2025h1/terragrunt.hcl
```diff
@@ -6,7 +6,7 @@ locals {
   env                                        = include.root.locals.env
   docker_env                                 = local.env
   base_name                                  = include.root.locals.base_name
-  origin_suffix                              = include.root.locals.origin_suffix
+  origin                                     = "${local.base_name}${include.root.locals.origin_suffix}"
```
phbnf: You can also edit / remove `origin_suffix` from `deployment/live/gcp/static-ct-staging/logs/root.hcl`

PR #697, opened 2026-02-09
Use cosigv1-formatted vkey for TF witness

AlCutter:
This PR updates the TF witness public vkey to one which explicitly specifies that it's an `0x04` Cosignature/v1 key. The public key material remains the same, though.

Since the Tessera witness support code will accept either `0x01` or `0x04` vkeys, this should be a `NOP`.


phbnf (approved):
What's the goal of differentiating Ed25519 note signatures and co-signatures? Should we modify https://github.com/transparency-dev/witness and/or github.com/transparency-dev/formats/log such that they can tell such signatures appart?

AlCutter:
> What's the goal of differentiating Ed25519 note signatures and co-signatures? Should we modify https://github.com/transparency-dev/witness and/or github.com/transparency-dev/formats/log such that they can tell such signatures appart?

They have a different pre-image and slightly different semantics as the latter requires the signer's view of time to be mixed in.  They can already be distinguished since the both the `vkey` and the signatures it verifies contain the `algoId` byte (`0x01` vs `0x04` in this case).

What's happening is that, mostly for legacy reasons, [`formats/note.NewVerifierForCosignatureV1()`](https://github.com/transparency-dev/formats/blob/main/note/note_cosigv1.go#L89) will internally "convert" a vanilla `Ed25519` vkey to a `Cosig/v1` key if you explicitly ask it to create a `Cosig/v1` verifier but pass in an `0x01` vkey.

Issue #416, opened 2025-07-10
Switch to structured loggin


phbnf:
TesseraCT uses klog and implements some abstraction with `requestLog`. Let's switch to structured logging, probably with slog, to have better logs. It should also make them more usable in Log Explorer. At the moment on GCP,  I have not found a way to merge logs of all VMs within a Managed Instance Group. I'm hoping that structured logs would help solve this?

PR #841, opened 2026-04-28
Switch all klog to slog

patflynn:
Migrates all logging call sites from k8s.io/klog/v2 to log/slog. Each binary now configures slog.SetDefault with a text handler keyed off a new --slog_level flag (the GCP server keeps its existing Cloud Logging exporter wiring). klog.V(N) call sites collapse to slog.Debug, and klog.Exitf/Fatalf are replaced with slog.Error + os.Exit(1).

Adds golangci-lint config enforcing sloglint (attr-only, context: all) and a depguard rule denying further klog imports.

Towards #416.

### DIFF
--- a/deployment/modules/aws/tesseract/conformance/main.tf
+++ b/deployment/modules/aws/tesseract/conformance/main.tf
@@ -178,7 +178,7 @@ resource "aws_ecs_task_definition" "conformance" {
       "--enable_publication_awaiter=true",
       formatlist("--roots_remote_fetch_url=%s", var.roots_remote_fetch_url),
       "--roots_remote_fetch_interval=${var.roots_remote_fetch_interval}",
-      "-v=2",
+      "--slog_level=-4",
     ]),
     "logConfiguration" : {
       "logDriver" : "awslogs",
@@ -258,12 +258,11 @@ resource "aws_ecs_task_definition" "hammer" {
       "--log_url=https://${module.storage.s3_bucket_regional_domain_name}",
       "--write_log_url=http://${aws_service_discovery_service.conformance.name}.${aws_service_discovery_private_dns_namespace.internal.name}:${local.port}/ci-static-ct",
       "--show_ui=false",
-      "--logtostderr",
       "--num_writers=256",
       "--max_write_ops=256",
       "--num_mmd_verifiers=1024",
       "--leaf_write_goal=10000",
-      "-v=3",
+      "--slog_level=-4",
     ],
     "logConfiguration" : {
       "logDriver" : "awslogs",
--- a/internal/ct/handlers.go
+++ b/internal/ct/handlers.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"math"
 	"math/rand/v2"
 	"net/http"
@@ -41,7 +42,6 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"golang.org/x/net/idna"
 	"golang.org/x/time/rate"
-	"k8s.io/klog/v2"
 )
 
 const (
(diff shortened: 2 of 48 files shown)


internal/ct/requestlog.go
phbnf, line 1: This file has its own structured logging format and should be migrated to slog as well. Quickly glancing at this PR, I think it does most of it for this file, maybe except for "RL"? Just flagging this such that whoever picks this up knows to pay attention to this one. I'm not even sure that we want to keep this request log as it is, I haven't looked at it in a while.

internal/ct/handlers.go
```diff
@@ -206,13 +206,13 @@ func (a appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Verify that the request was received at an URL starting with the origin, as per https://c2sp.org/static-ct-api.
 	// Don't block requests that don't satisfy this to allow for custom proxy configuration, or custom request routing.
 	if err := receivedAtOrigin(r, a.log.origin); err != nil {
-		klog.Warningf("%s: %s the request was received on a URL which is not prefixed with the configured origin: %v", a.log.origin, a.name, err)
+		slog.WarnContext(logCtx, "the request was received on a URL which is not prefixed with the configured origin", slog.String("origin", a.log.origin), slog.String("name", a.name), slog.Any("error", err))
 	}
 
-	klog.V(2).Infof("%s: request %v %q => %s", a.log.origin, r.Method, r.URL, a.name)
+	slog.DebugContext(logCtx, "request received", slog.String("origin", a.log.origin), slog.String("method", r.Method), slog.String("url", r.URL.String()), slog.String("name", a.name))
```
mhutchinson, line 212: This will be extremely noisy for a public-facing log if debug is enabled, but I guess that's the name of the game. I think it's orthogonal to this PR, but a bunch of this logging should be reviewed to see if it's still needed. Perhaps traces and events are a better way of mapping out the "what happened on this request?" story.

mhutchinson (approved):

roger2hk (commented):
Note that there is a difference between `klog.Exitf` and the combination of `slog.ErrorContext` and `os.Exit(1)`. The new behavior will not flush the logs before exiting. For those async logging, there is a chance to lose the final log.

internal/ct/requestlog.go
patflynn: Filed #842 to track this — same conclusion as you reached: the `RL:` prefix is preserved verbatim in the migration, but it's worth a separate look at whether to keep, refactor, or drop the per-request log channel entirely.

internal/ct/handlers.go
patflynn: Agreed this is orthogonal to the migration — filed #843 to track the audit (drop / fold into OTel spans / keep with justification).

patflynn:
@roger2hk good catch on the flush behaviour — addressed in 6c7be3e. The GCP binary now has a `logFlushers` slice that `initLogging` populates with the Cloud Logging client's `Close()`, and a `fatal()` helper that drains them before `os.Exit(1)`. All `slog.ErrorContext + os.Exit(1)` callsites in `cmd/tesseract/gcp/{main,otel}.go` now go through `fatal()`.

The text-handler-based binaries (aws, posix, fsck, hammer, migrate, fetch_roots) write synchronously to stderr, so they don't need the same treatment.

internal/ct/handlers.go
phbnf: Before this PR, we could control the verbosity of `klog.V(X)` by setting `-v=X`, and most binaries were configured with `-v=1`.

All those entries have been migrated to `slog.DebugContext`, and `-v=X` flags to `slog_level=-4`. That means that if this PR is rolled out, the number of logging entries will be much (much) higher. Specifically, all the `requestlog.go` logs had a -v level of 9 and had always been disabled, now these entries are indistinguishable from the ones that had a -v level of 1.

We've observed performance regression because of logging in the past (https://github.com/transparency-dev/tesseract/issues/756), so we should be mindful of this. I don't know what the impact is now that we've migrated to `slog`, but that's one of the thing that we need to keep an eye on as part of landing the slog work.

deployment/modules/aws/tesseract/conformance/main.tf
```diff
@@ -178,7 +178,7 @@ resource "aws_ecs_task_definition" "conformance" {
       "--enable_publication_awaiter=true",
       formatlist("--roots_remote_fetch_url=%s", var.roots_remote_fetch_url),
       "--roots_remote_fetch_interval=${var.roots_remote_fetch_interval}",
-      "-v=2",
+      "--slog_level=-4",
```
phbnf, line 181: This needs to be rolled out, the conformance test is broken in the mean time, I'm on it.

phbnf: Done!

internal/ct/handlers.go
phbnf: I rolled this out to the conformance pipeline, and this resulted in a ~5-10X of logging entries: not rocket science but quickly look at logs it went from ~30k entries to ~210k.

Issue #884, opened 2026-06-24
Improve robustness when constructing SCT for duplicate entry


AGWA:
Following up from https://github.com/transparency-dev/tessera/issues/1018

I think it's fragile how tesseract overwrites just the timestamp, but otherwise keeps the entry that was constructed from the submission. It means that inputs to the critical SCT signing operation come from multiple places, and consequentially might not be coherent. Instead, tesseract should retrieve the entire entry from storage, not just the timestamp, and completely discard the entry constructed from the submission. That way, a malfunction in the antispam driver can only cause the wrong SCT to be returned; it can't kill the log by returning an invalid SCT.

phbnf:
Good point, thank you very much for the report @AGWA. I think we should also check that all the fields returned from deduplication match with the submission, not to return any erroneous SCT.

phbnf:
Fixed now, thank you very much @AGWA  for reporting this.

As you suggested, for duplicate requests, the SCT input is now built based on the entry already available in storage.

On top of this:
 - TesseraCT checks that this SCT input matches with the request it received.
 - I also added some SCT verification so the `hammer/` tool which we use in our ci tests to increases chances of detecting any discrepancies (although there shouldn't be any).

PR #886, opened 2026-06-26
More robust SCT creation for duplicate entries

phbnf:
Towards #884 

This PR modifies builds SCTs from the leaf data of deduplicated entries, rather than the submitted entry. These should always match, but this defensive change prevents TesseraCT from signing an SCT that would not correspond to a leaf.

This PR also:
 - checks that the corresponding SCT matches with the submitted entry
 - simplifies function signatures, to only return the required data

### DIFF
--- a/internal/ct/handlers.go
+++ b/internal/ct/handlers.go
@@ -17,6 +17,7 @@ package ct
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
@@ -33,6 +34,7 @@ import (
 	"time"
 
 	"github.com/transparency-dev/tessera"
+	"github.com/transparency-dev/tessera/ctonly"
 	"github.com/transparency-dev/tesseract/internal/logger"
 	"github.com/transparency-dev/tesseract/internal/otel"
 
@@ -512,10 +514,14 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
 			w.Header().Add("Retry-After", strconv.Itoa(rand.IntN(5)+1)) // random retry within [1,6) seconds
 			return http.StatusTooManyRequests, []attribute.KeyValue{duplicateKey.Bool(index.IsDup), tooManyRequestsReasonKey.String("rate_limit_dedup")}, errors.New(http.StatusText(http.StatusTooManyRequests))
 		}
-		entry.Timestamp, err = log.storage.DedupFuture(ctx, future)
+		dupEntry, err := log.storage.DedupFuture(ctx, future)
 		if err != nil {
 			return http.StatusInternalServerError, []attribute.KeyValue{duplicateKey.Bool(index.IsDup)}, fmt.Errorf("could not resolve duplicate: %v", err)
 		}
+		if err := entryMatches(dupEntry, entry); err != nil {
+			return http.StatusInternalServerError, []attribute.KeyValue{duplicateKey.Bool(index.IsDup)}, fmt.Errorf("deduplicated entry in storage does not match submitted entry: %w", err)
+		}
+		entry = dupEntry
 	}
 
 	// Always use the returned leaf as the basis for an SCT.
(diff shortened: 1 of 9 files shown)


internal/ct/handlers.go
```diff
@@ -507,29 +510,29 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
...
 			return http.StatusTooManyRequests, []attribute.KeyValue{duplicateKey.Bool(index.IsDup), tooManyRequestsReasonKey.String("rate_limit_dedup")}, errors.New(http.StatusText(http.StatusTooManyRequests))
 		}
-		entry.Timestamp, err = log.storage.DedupFuture(ctx, future)
+		var err error
+		sctInput, err = log.storage.DedupFuture(ctx, future)
 		if err != nil {
 			return http.StatusInternalServerError, []attribute.KeyValue{duplicateKey.Bool(index.IsDup)}, fmt.Errorf("could not resolve duplicate: %v", err)
 		}
+		if err := entrySCTsMatch(sctInput, entry, index.Index); err != nil {
+			return http.StatusInternalServerError, []attribute.KeyValue{duplicateKey.Bool(index.IsDup)}, fmt.Errorf("deduplicated entry in storage does not match submitted entry: %w", err)
```
AlCutter: Do you need the `%w` for the caller behave correctly, or is `%v` enough?

internal/ct/handlers.go
```diff
@@ -553,6 +556,33 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
 	return http.StatusOK, []attribute.KeyValue{duplicateKey.Bool(index.IsDup)}, nil
 }
 
+// entrySCTsMatch checks that sctInput fields match with an entry.
+// It checks for all the sctInput fields, except for the timestamp which might
+// not be set in the entry yet.
+func entrySCTsMatch(sct *rfc6962.CertificateTimestamp, e *ctonly.Entry, idx uint64) error {
+	extractedIdx, err := staticct.ParseCTExtensionsBytes(sct.Extensions)
+	if err != nil || extractedIdx != idx {
+		return fmt.Errorf("index mismatch: stored=%d, expected=%d", extractedIdx, idx)
```
AlCutter, line 569: This message seems like it could be misleading if `err != nil`?

internal/types/staticct/staticct.go
```diff
@@ -192,11 +264,21 @@ func ParseCTExtensions(ext string) (uint64, error) {
...
 		!extensions.Empty() {
-		return 0, fmt.Errorf("invalid SCT extension data: %v", ext)
+		return 0, fmt.Errorf("invalid SCT extension data: %x", ext)
 	}
 	return leafIdx, nil
 }
 
+// ParseCTExtensions parses base64-encoded CTExtensions into an index.
+// Code is inspired by https://github.com/FiloSottile/sunlight/blob/main/tile.go.
+func ParseCTExtensions(ext string) (uint64, error) {
```
AlCutter: nit: could rename to `ParseCTExtensionsB64`? Looks like the only place this is used is in the hammer (could also just make the hammer do the de-b64)

internal/ct/handlers.go
```diff
@@ -556,12 +557,21 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
...
 // It checks for all the sctInput fields, except for the timestamp which might
 // not be set in the entry yet.
-func entrySCTsMatch(sct *rfc6962.CertificateTimestamp, e *ctonly.Entry, idx uint64) error {
+func sctMatchesEntry(sct *rfc6962.CertificateTimestamp, e *ctonly.Entry, idx uint64) error {
+	if sct == nil {
+		return errors.New("sct is nil")
+	}
+	if e == nil {
+		return errors.New("entry is nil")
+	}
```
AlCutter: Wondering if you can avoid all these `nil` checks  here (and in `Sign` below)?

Is it ever acceptable for `sctInput == nil && err == nil` at L524 or for `entry` at L477?

Could the funcs which return them return by-value?

(Can totally be a follow-up PR!)

AlCutter (approved):

internal/ct/handlers.go
phbnf: We should never run into these edge cases with the current implementation, but we can't tell for sure what will call this method later, so I addd these defensive checks. I considered passing structs by value but ruled it out to avoid large copies since the cert is in there. After looking into it again, certs wouldn't be copied, only the reference to the byte array. Since copies will be small, it shouldn't be a problem. I've changed the signature of `signSCT` and `DedupFuture`. We can dereference the pointer for `ctonly.Entry`. `EntryFromChain` uses `sync.Pool`, so I'd apply extra care before returning a struct rather than a pointer.

I've left checks for `sct.PrecertEntry` and `sct.X509Entry`. They are not necessary with the current code paths, but better err on the safe side.

AlCutter (approved):

PR #894, opened 2026-07-07
Log error at error level for 500 errors

cmurphy:
An HTTP 500 should be logged as an error since an operator likely needs
to take action to fix the problem.

### DIFF
--- a/internal/ct/handlers.go
+++ b/internal/ct/handlers.go
@@ -246,7 +246,11 @@ func (a appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			statusCode = ClientClosedRequestStatus
 			err = fmt.Errorf("client closed the connection: %v", err)
 		}
-		slog.WarnContext(ctx, "handler error", slog.String("origin", a.log.origin), slog.String("name", a.name), slog.Any("error", err))
+		if statusCode == http.StatusInternalServerError {
+			slog.ErrorContext(ctx, "handler error", slog.String("origin", a.log.origin), slog.String("name", a.name), slog.Any("error", err))
+		} else {
+			slog.WarnContext(ctx, "handler error", slog.String("origin", a.log.origin), slog.String("name", a.name), slog.Any("error", err))
+		}
 		a.opts.sendHTTPError(w, statusCode, err)
 		span.RecordError(err)
 		span.SetStatus(codes.Error, err.Error())


cmurphy:
See also https://github.com/sigstore/fulcio/pull/2393

internal/ct/handlers.go
```diff
@@ -522,26 +526,31 @@ func addChainInternal(ctx context.Context, opts *HandlerOptions, log *log, w htt
...
 	if err != nil {
+		slog.ErrorContext(ctx, "Failed to marshall SCT", slog.Any("error", err))
 		return http.StatusInternalServerError, nil, fmt.Errorf("failed to marshall SCT: %s", err)
 	}
 	// We could possibly fail to issue the SCT after this but it's v. unlikely.
 	opts.RequestLog.issueSCT(ctx, sctBytes)
 	err = marshalAndWriteAddChainResponse(sct, w)
 	if err != nil {
 		// reason is logged and http status is already set
+		slog.ErrorContext(ctx, "Failed to write response", slog.Any("error", err))
```
AlCutter: I'd be tempted to remove this one. All(?) the others are rare and indicative of something local going wrong (either local infra, or bugs).

This one, I suspect, is almost always just going to be a badly behaved client closing their socket early and so likely not actionable.

AlCutter (approved):
Most of these seem like a good call - they should be rare but you'd want to know about them.

phbnf:
I think that error messages should already propagate to [warnings on line 249](https://github.com/transparency-dev/tesseract/blob/af2278a632ede926605bf31999514baabc135322/internal/ct/handlers.go#L249). Is there a scenario in which they wouldn't?

cmurphy:
> I think that error messages should already propagate to warnings on line 249. Is there a scenario in which they wouldn't?

We are seeing 500 errors frequently with no logs at any level to indicate what's going on.

cmurphy:
> We are seeing 500 errors frequently with no logs at any level to indicate what's going on.

When I searched explicitly for "handler error" at the warning level I finally found it, it had been lost in a sea of other warnings. Since this caused a 500 I think this should be logged at the Error level rather than Warning so that it would be easier to find.

cmurphy:
I've changed this to not log within the individual handler but to log at an error level when the handler returns.

phbnf (approved):

PR #897, opened 2026-07-10
Add client-side SCT verification and entry matching in hammer

### LINKED ISSUE #884: Improve robustness when constructing SCT for duplicate entry

phbnf:
This PR allows the hammer to verify that the SCTs it receives matches with the requests it sent.

While I'm there, move `isPrecertificate` to `x509util`.

I tested this with the hammer locally and it worked. I'll send a PR afterwards that adds instructions to run the hammer locally with a POSIX log unless someone can point me at them!

Fixes #884.

### DIFF
--- a/internal/hammer/hammer.go
+++ b/internal/hammer/hammer.go
@@ -18,10 +18,12 @@ package main
 import (
 	"bytes"
 	"context"
+	"crypto"
 	"crypto/ecdh"
 	"crypto/ecdsa"
 	"crypto/ed25519"
 	"crypto/rsa"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
@@ -49,6 +51,8 @@ import (
 	"github.com/transparency-dev/tesseract/internal/hammer/loadtest"
 	"github.com/transparency-dev/tesseract/internal/types/rfc6962"
 	"github.com/transparency-dev/tesseract/internal/types/staticct"
+	tlstypes "github.com/transparency-dev/tesseract/internal/types/tls"
+	"github.com/transparency-dev/tesseract/internal/x509util"
 	"golang.org/x/mod/sumdb/note"
 	"golang.org/x/net/http2"
 )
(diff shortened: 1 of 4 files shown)


phbnf:
Gentle ping on that one, I'd like to get it in to close on #884.

internal/hammer/hammer.go
```diff
@@ -544,18 +562,107 @@ func retryDelay(retryAfter string, defaultDur time.Duration) time.Duration {
...
+		return func(reqBytes []byte, respBytes []byte) (uint64, uint64, error) {
+			return 0, 0, errors.New("cannot verify SCT: log public key is nil")
+		}
 	}
-
-	leafIdx, err := staticct.ParseCTExtensionsB64(resp.Extensions)
+	pubBytes, err := x509.MarshalPKIXPublicKey(pubKey)
 	if err != nil {
-		return 0, 0, fmt.Errorf("can't parse extensions: %v", err)
+		return func(reqBytes []byte, respBytes []byte) (uint64, uint64, error) {
```
AlCutter, line 583: It feels like at this point we know the user intent is to verify (they've provided a "pub key"), but the key is broken so we should bail rather than return a func which will always error out.

Could perhaps return `(func..., error)`, and just check the err from main rather than inlining in a chained call?

internal/hammer/hammer.go
```diff
@@ -544,18 +562,107 @@ func retryDelay(retryAfter string, defaultDur time.Duration) time.Duration {
...
+
+		switch pk := pubKey.(type) {
+		case *ecdsa.PublicKey:
+			if !ecdsa.VerifyASN1(pk, digest[:], ds.Signature) {
+				return 0, 0, errors.New("SCT ECDSA signature verification failed")
+			}
+		case *rsa.PublicKey:
+			if err := rsa.VerifyPKCS1v15(pk, crypto.SHA256, digest[:], ds.Signature); err != nil {
+				return 0, 0, fmt.Errorf("SCT RSA signature verification failed: %w", err)
+			}
```
AlCutter: can probably safely ditch this RSA support and rely on the `default` "computer says no"?

Pretty sure there are no RSA static CT logs and never should there be :)

AlCutter (approved):

roger2hk (approved):

internal/hammer/hammer.go
phbnf: Nice, thanks! I've also added the same check when the key is nil, at which point, you might as well not call the function if your intent it actually to have a nil key.

PR #932, opened 2026-09-02
Bugfix hunt

robstradling:
I asked Claude to "Hunt for any obvious bugs, and suggest fixes."


roger2hk (approved):
Thanks @robstradling for the fix. The bug in `internal/types/tls/tls.go` can also be found in ct-go repo.

https://github.com/google/certificate-transparency-go/blob/master/tls/tls.go#L339

mhutchinson (approved):

PR #945, opened 2026-09-16
Migrate GCP OTel exporters to OTLP

phbnf:
The GoogleCloudPlatform/opentelemetry-operations-go metric and trace exporters are deprecated. Replace them with the standard OTLP gRPC exporters pointed at telemetry.googleapis.com, as in transparency-dev/tessera#1169. Unblocks #944.


AlCutter (approved):
Note that the metric names change with this new exporter: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/MIGRATION.md#migrate-from-google-cloud-monitoring-exporter-to-otlp-exporter

phbnf:
Thanks. That's quite a change, and it should definitely be mentioned in release notes. I'll submit this PR in the interest of moving forward, but should we introduce a `CHANGELOG.md` file like we have in other repos to make sure we don't drop such changes from release notes?

AlCutter:
> Thanks. That's quite a change, and it should definitely be mentioned in release notes. I'll submit this PR in the interest of moving forward, but should we introduce a `CHANGELOG.md` file like we have in other repos to make sure we don't drop such changes from release notes?

Yeah, that sounds like a good idea!

